PR #7, opened 2016-06-20
Use redis for local and locking memcache.

### LINKED ISSUE #2: Enable memcache
Upon visiting the admin page after initial install, says that there is no memcache installed.

kyrofa:
This PR resolves #2 by using redis for both the local and locking memcache. It runs via a unix socket only.


oparoz:
Had time to review this and LGTM

PR #23, opened 2016-08-17
Add support for HTTPS.

### LINKED ISSUE #6: Enable HTTPS
Upon visiting the admin page after initial install, it recommends the set-up of HTTPS.

kyrofa:
This PR resolves #6 by accomplishing the following:

- Nothing changes upon first install-- HTTP only.
- Add an `enable-https` command for either generating a self-signed certificate or obtaining one from Let's Encrypt. If used, the command changes the snap to HTTPS-only.
- Add `disable-https` command to revert back to HTTP-only.

Please test this version of the snap with the following (amd64 only, please let me know if you need another arch):

```bash
$ sudo snap install nextcloud --beta
```

The initial experience should be the same as it currently is: HTTP only. Now let's enable HTTPS.

```bash
$ /snap/bin/nextcloud.enable-https -h
Usage:
    /snap/nextcloud/x1/bin/enable-https [-h -t -d -s]

    -t: Obtain a test certificate. This is a valid Let's
        Encrypt certificate, but is not signed by a
        recognized CA, so browsers won't accept it.
    -d: Dry run: don't actually obtain/install certificates
        from Let's Encrypt, but make sure it's possible.
    -s: Generate and use a self-signed certificate. This
        is easier to setup than Let's Encrypt certificates,
        but will cause warnings in browsers.
```

Let's start with a self-signed certificate, since it requires no setup:

```bash
$ sudo /snap/bin/nextcloud.enable-https -s
Generating key and self-signed certificate... done
Stopping apache... done
Starting apache... done
```

You should see that HTTP now redirects to HTTPS:

```bash
$ curl -s -vv -k http://xenial-vbx.local 2>&1 | grep "< "
< HTTP/1.1 301 Moved Permanently
< Date: Sat, 20 Aug 2016 20:07:51 GMT
< Server: Apache/2.4.23 (Unix) OpenSSL/1.0.2g-fips
< Location: https://xenial-vbx.local/
< Content-Length: 233
< Content-Type: text/html; charset=iso-8859-1
```

Also note that HSTS is not enabled (since the certificate is self-signed):

```bash
$ curl -s -vv -k https://xenial-vbx.local 2>&1 | grep "Strict"
```

Alright, disable HTTPS:

```bash
$ sudo /snap/bin/nextcloud.disable-https 
Deactivating HTTPS... done
Stopping apache... done
Starting apache... done
```

Note that it's now HTTP-only:

```bash
$ curl -k https://xenial-vbx.local
curl: (7) Failed to connect to xenial-vbx.local port 443: Connection refused
```

Now let's obtain a certificate from Let's Encrypt:

```bash
$ sudo /snap/bin/nextcloud.enable-https
In order for Let's Encrypt to verify that you actually own the domain for
which you're requesting a certificate, there are a number of requirements
of which you need to be aware:

1. In order to register with the Let's Encrypt ACME server, you must agree
   to the currently-in-effect Subscriber Agreement located here:

       https://letsencrypt.org/repository/

   By continuing to use this tool you agree to these terms. Please cancel
   now if otherwise.

2. You must have the domain name(s) for which you want certificates
   pointing at the external IP address of this machine.

3. Both ports 80 and 443 on the external IP address of this machine must
   point to this machine (e.g. port forwarding might need to be setup on
   your router).

Have you met these requirements? (y/n) y
Please enter an email address (for urgent notices or key recovery): foo@foo.com
Please enter your domain name(s) (space-separated): my.domain.com
Attempting to obtain certificates... done
Stopping apache... done
Starting apache... done
```

HTTP should again be redirecting to HTTPS, and HSTS should be enabled:

```bash
$ curl -vv -k https://my.domain.com 2>&1 | grep "Strict"
< Strict-Transport-Security: max-age=63072000; includeSubdomains;
```

As a result, this snap should now obtain an A+ rating on [SSLLabs](https://www.ssllabs.com/ssltest/). Note that the certificate will automatically be updated for you.


shasderias:
Tested this. Worked as expected. OOB experience was very positive.

FWIW, prefer the position you have taken (i.e. HTTPS only, self-signed cert by default). HTTPS with a warning would still be preferable to bare HTTP.

Let me know if you want any specific feedback.

kyrofa:
> Tested this. Worked as expected. OOB experience was very positive.

I really appreciate your testing this out, thank you!

> FWIW, prefer the position you have taken (i.e. HTTPS only, self-signed cert by default). HTTPS with a warning would still be preferable to bare HTTP.

If this was as far as we were going, I'd agree with you. The process of enabling HTTPS is too hard right now not to make it the default. However, @jospoortvliet, @oparoz and I met yesterday about this, and discussed it in terms of longer-term plans:

1. Nextcloud warns about the use of HTTP on the admin page so hopefully people will know that's not the ideal setup.
2. We plan on making a Nextcloud app to enable HTTPS rather than requiring one to SSH in and run a script (in a later release).
3. We could even make the initial pop-up contain some information about how (and why) to enable HTTPS.

Also, the Nextcloud snap is already installed by plenty of people and they'll automatically get this update. We don't want them to all of a sudden start getting HTTPS warnings. What if they don't even have port 443 open in their firewall or router? All of a sudden this snap is useless to them without some manual work.

All of these things lead to our wanting to change this PR slightly:

1. Use HTTP-only out of the box.
2. Update the `setup-https` script to support either self-signed or Let's Encrypt certificates. If run, make it HTTPS-only using the certificate type selected.

kyrofa:
@oparoz @jospoortvliet alright, this works the way I think we want it to after our meeting. Please review when you're able.

snapcraft.yaml
```diff
@@ -66,18 +82,40 @@ parts:
...
+  apache-customizations:
+    plugin: copy
+    files:
+      src/apache/scripts/*: bin/
+      src/apache/conf/*: conf/
+
+  nextcloud:
+    plugin: copy
+    source: https://github.com/kyrofa/nextcloud.git
+    source-tag: 9.0.50
```
oparoz: We're at 9.0.53 ;)

kyrofa: :disappointed: thanks, this feature obviously began before the upgrade, looks like it survived the rebase. Fixed!

thomascobb:
Hi, I'd love to try this out, but I'm on a 32-bit system. If you still need people to test and are willing to make an i686 package then I'll have a go...

kyrofa:
> Hi, I'd love to try this out, but I'm on a 32-bit system. If you still need people to test and are willing to make an i686 package then I'll have a go...

Thanks for chiming in @thomascobb. I've now uploaded a build of this PR to all architectures on the beta channel, so please give it a shot!

thomascobb:
Thanks, I'll give a go at the weekend

thomascobb:
Cool, I ran nextcloud.enable-https -t and I now have https working as expected. However, I get T (If trust issues are ignored: A) rather than A+ on SSL labs. Is that to be expected?

kyrofa:
> Cool, I ran nextcloud.enable-https -t and I now have https working as expected. However, I get T (If trust issues are ignored: A) rather than A+ on SSL labs. Is that to be expected?

Yes, this is expected. the T is due to the fact that your certificate is not signed by a trusted third party (it's signed by Let's Encrypt's staging server, since you used the `-t` flag). If you want an A+, you need to use `nextcloud.enable-https` with no parameters, which will obtain a real Let's Encrypt certificate.

PR #34, opened 2016-08-29
Increase PHP and Apache proxy timeout to 3600 seconds.

kyrofa:
Image verification noted a failure to upload 1.5GB file on rpi2, logs indicated a timeout occurred.


oparoz:
AFAIK, this needs to match what's in the Apache config. Is it also set at 3600?

kyrofa:
> AFAIK, this needs to match what's in the Apache config. Is it also set at 3600?

To which parameter in the Apache config are you referring?

oparoz:
https://github.com/nextcloud/nextcloud-snap/blob/master/src/apache/conf/httpd.conf#L149

kyrofa:
Good catch @oparoz, thank you!

PR #41, opened 2016-09-10
Updated help text for enable-https script

thomascobb:
I missed the fact that you could run this script without arguments to get a Let's Encrypt certificate. Hopefully this bit of text will mean others don't make the same mistake...


kyrofa:
Thanks @thomascobb! Apparently I assumed that indicating the options were optional (with the `[]`) would magically describe to the user what the script did with no options :stuck_out_tongue: .

kyrofa:
Please take a quick look at the [contribution guide](https://github.com/nextcloud/nextcloud-snap/blob/master/CONTRIBUTING.md) when you're able. In summary:

1. Please log an issue for this.
2. Base your work (and make a PR back into) **develop**, not master.
3. Mention the bug you created in your commit in such a way as to automatically close it (not in the title, please-- just on a line all by itself).

We have a CI system that runs and produces snaps automatically depending on the branch being merged into. Merging to master would spin a new stable release, and we're not ready to do that yet. Merging to develop will spin a new beta release (though soon it will be RC instead). So I'll close this for now-- please re-proposed to develop when you're able.

PR #42, opened 2016-09-11
Added service that runs cron.php every 15 minutes

thomascobb:
Not sure if I'm doing something wrong, but I had to restart the computer before the service would start. I tried to recreate it, but it always seemed to start correctly.

I checked dmesg, and I seem to have a lot of messages like this:

[  845.275631] audit: type=1400 audit(1473627189.253:332): apparmor="DENIED" operation="exec" profile="snap.nextcloud.mysql" name="/bin/run-parts" pid=6828 comm="mysql.server" requested_mask="x" denied_mask="x" fsuid=0 ouid=0
[  845.500563] audit: type=1400 audit(1473627189.481:333): apparmor="DENIED" operation="ptrace" profile="snap.nextcloud.mysql" pid=6982 comm="ps" requested_mask="trace" denied_mask="trace" peer="unconfined"
[  845.507404] audit: type=1400 audit(1473627189.485:334): apparmor="DENIED" operation="ptrace" profile="snap.nextcloud.mysql" pid=6982 comm="ps" requested_mask="trace" denied_mask="trace" peer="unconfined"
[  845.507491] audit: type=1400 audit(1473627189.485:335): apparmor="DENIED" operation="ptrace" profile="snap.nextcloud.mysql" pid=6982 comm="ps" requested_mask="trace" denied_mask="trace" peer="unconfined"
[  845.507569] audit: type=1400 audit(1473627189.485:336): apparmor="DENIED" operation="ptrace" profile="snap.nextcloud.mysql" pid=6982 comm="ps" requested_mask="trace" denied_mask="trace" peer="unconfined"
[  845.508631] audit: type=1400 audit(1473627189.489:337): apparmor="DENIED" operation="ptrace" profile="snap.nextcloud.mysql" pid=6982 comm="ps" requested_mask="trace" denied_mask="trace" peer="unconfined"
[  845.508710] audit: type=1400 audit(1473627189.489:338): apparmor="DENIED" operation="ptrace" profile="snap.nextcloud.mysql" pid=6982 comm="ps" requested_mask="trace" denied_mask="trace" peer="unconfined"
[  845.508934] audit: type=1400 audit(1473627189.489:339): apparmor="DENIED" operation="ptrace" profile="snap.nextcloud.mysql" pid=6982 comm="ps" requested_mask="trace" denied_mask="trace" peer="unconfined"
[...]

PR #45, opened 2016-09-13
Add service that runs cron.php every 15 minutes

### LINKED ISSUE #39: Use cron rather than Ajax cron
I'm new to nextcloud, but my understanding is that if you're not going to visit the webpage regularly then you should use cron or web cron rather than Ajax cron. I am mainly planning on using the desktop client for sync so I think this applies to me? I had a look at how you've done the renew certificates, and I think a similar approach running cron.php with 15 minute sleeps might be a good idea? Do you think this would work? I could try to do this and make a pull request if that's helpful.

thomascobb:
Fixes #39 

A second attempt based on #42 

I changed the name of the service to nextcloud-cron as suggested, and moved the script into src/nextcloud/scripts/nextcloud-cron as that seemed to make more sense.

When building, I needed to do:
snapcraft clean
snapcraft

So that the wildcard in nextcloud-customizations picked up the new nextcloud-cron script...

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -66,6 +66,12 @@ apps:
     restart-condition: always
     plugs: [network, network-bind]
 
+  nextcloud-cron:
+    command: nextcloud-cron
+    daemon: simple
+    restart-condition: always
+    plugs: [network, network-bind]
+
 parts:
   apache:
     plugin: apache
--- a/src/nextcloud/scripts/nextcloud-cron
+++ b/src/nextcloud/scripts/nextcloud-cron
@@ -1,0 +1,15 @@
+#!/bin/sh
+
+export NEXTCLOUD_CONFIG_DIR=$SNAP_DATA/nextcloud/config
+
+echo -n "Waiting for Nextcloud config dir... "
+while [ ! -d "$NEXTCLOUD_CONFIG_DIR" ]; do
+	sleep 1
+done
+
+echo "done"
+
+while true; do
+	php -c $SNAP/config/php $SNAP/htdocs/cron.php
+	sleep 15m
+done


src/nextcloud/scripts/nextcloud-cron
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+export NEXTCLOUD_CONFIG_DIR=$SNAP_DATA/nextcloud/config
```
kyrofa: The reason you ran into issues at first is because upon first install it takes a little while for this directory to exist. In order to avoid what looks like an error (but isn't), I suggest before starting the cron loop below, you sit and wait for this directory to exist. Something like the following:

echo -n "Waiting for Nextcloud config dir... "
while [ ! -d "$NEXTCLOUD_CONFIG_DIR" ]; then
    sleep 1
done
echo "done"

snapcraft.yaml
```diff
@@ -66,6 +66,11 @@ apps:
     restart-condition: always
     plugs: [network, network-bind]
 
+  nextcloud-cron:
+    command: nextcloud-cron
+    daemon: simple
+    restart-condition: always
```
kyrofa: This actually requires the network, so please include `plugs: [network, network-bind]` here. Otherwise it's killed by seccomp.

thomascobb:
Your suggestions have been implemented, and now it works both from a fresh install and as an update from the existing released snap. Each time I did the snap install, then set the cron setting in the admin page, then waited 15 minutes and verified that the timer on the admin page had been reset.

fabwu:
@thomascobb Is this the correct cron mode?

image

If so I add it to the documentation so people can apply it to their installation.

thomascobb:
It's the third option, system cron

PR #63, opened 2016-09-27
Tighten SSL config

### LINKED ISSUE #51: Tighten default SSL config
## Tighten default config

There isn't much to do here, because we need to keep backward compatibility with TLS1.0. It's bad, but we can't do anything about it until the OSX client is updated.

### Changes

* Remove DES3
* 384bits curve only
* Use 2048 bits DH parameters

```
SSLCipherSuite ECDHE+AESGCM:DHE+AESGCM:ECDHE+AES256:DHE+AES256:ECDHE+AES128:DHE+AES:RSA+AESGCM:RSA+AES:!aNULL:!3DES:!MD5:!DSS;
SSLOpenSSLConfCmd DHParameters "${SNAP}/conf/dhparams.2048.pem
```

oparoz:
This PR only alters the default configuration. We can't make it safe for all users as the OSX client requires TLS 1.0 compatibility.
User who need to a safe configuration can find it in `/var/snap/nextcloud/current/apache/conf/custom.conf`

Fixes #51

Tested on Chrome, Firefox, Edge, IE11

## Todo

- [ ] Will be rebased once #71 is merged

## New default config

default_config


kyrofa:
As I mentioned when we were discussing the bug, I've seen no information about why this is an improvement over the current config. I'd like to see some objective data for it. May I suggest comparing reports from https://www.ssllabs.com/ssltest/ ?

oparoz:
I did post a link which explains DH parameters in the original issue and the rest is just hosting industry best practices. You can find an example recommendation on Mozilla's website:
https://wiki.mozilla.org/Security/Server_Side_TLS
and this is just an implementation of that recommendation. I've just not added chacha20 since we don't have the OpenSSL for it yet.

If the current settings get you an A+ on SSLlabs, you would think that they're OK, but they're not. SHA1 should be banned.

The settings I've modified could be tightened even more to achieve a full 100% on SSL labs, but that would mean using 4096bits DH parameters which, today, does nothing more than slow down connections and use more resources. AES128 ciphers would also need to be removed, but some could argue that these should even come first in the list. Advanced users will configure the order like they want to once we support custom configs.

oparoz:
## SSL labs rating with this config

ssllabs_a

ssllabs_proto

oparoz:
Just noticed a contradiction actually... I've removed the DHE ciphers, so the DHE params are not even needed... Will push a fix.

kyrofa:
> I did post a link which explains DH parameters in the original issue and the rest is just hosting industry best practices.

"hosting industry best practices" is not objective. Where do you think I got the original cipher suite?

> If the current settings get you an A+ on SSLlabs, you would think that they're OK, but they're not. SHA1 should be banned.

Forgive my ignorance, but isn't SHA1 a signature algorithm some CAs use on the certs? Let's Encrypt doesn't, and it's not something we control in the cipher suite unless I'm mistaken, so I'm not sure how this PR improves things in that regard.

src/apache/conf/ssl.conf
```diff
@@ -40,8 +40,7 @@ LoadModule ssl_module modules/mod_ssl.so
 #   Disable SSLv3 by default (cf. RFC 7525 3.1.1).  TLSv1 (1.0) should be
 #   disabled as quickly as practical.  By the end of 2016, only the TLSv1.2
 #   protocol or later should remain in use.
-SSLProtocol all -SSLv3
-SSLProxyProtocol all -SSLv3
+SSLProtocol All -SSLv3 -TLSv1 -TLSv1.1
```
kyrofa: The only difference in the ssllabs score between the current config and this PR is the protocol support, which I assume it due to this line. I bet if this is the only change you make you'd get the same score you just got. Unless there's a reason to change the suite, I say leave the rest of the config alone and only make this change.

oparoz:
> Where do you think I got the original cipher suite?

Well, I don't know :D , but I reckon that because ECDSA wasn't in your list, that it came from a decent source.

> Forgive my ignorance, but isn't SHA1 a signature algorithm some CAs use on the certs?

SHA1 has been deprecated for 5 years now. This article from Google is from 2 years ago:
https://security.googleblog.com/2014/09/gradually-sunsetting-sha-1.html

In any case, you won't be able to use Chrome and SHA1 in a couple of months.
So, that one is really safe to remove.
As I said earlier, it's not so black and white for AES128, so I think it's best to leave it in, even if it lowers the score.

src/apache/conf/ssl.conf
oparoz: Well, there is a reason to change the suite: Get rid of SHA1.

oparoz:
Since the certificate uses SHA256, then there is no point in having SHA1 in the list of ciphers

kyrofa:
Ah, I see where you're coming from now. I think you're confusing SHA1's use as a signature algorithm (which has been deprecated) with its use in the cipher suite (which has not). Perhaps [this](https://serverfault.com/questions/787358/apache-ssl-fs-disable-sha1) is pertinent?

kyrofa:
I'd be curious to see the complete SSL labs report. Client compatibility is also important.

oparoz:
> Ah, I see where you're coming from now. I think you're confusing SHA1's use as a signature algorithm (which has been deprecated) with its use in the cipher suite (which has not). Perhaps this is pertinent?

Ah, thanks for the link. It's true that I assumed SHA1 should simply not be used anywhere, without looking at the details, since removing it has had no impact.

oparoz:
I was following advice from:
https://bettercrypto.org/static/applied-crypto-hardening.pdf

Which has dropped SHA1 from their Configuration A, but they do have a note about HMAC-SHA1.

kyrofa:
Current SSL labs report for comparison:

original_ssllabs_test

oparoz:
I've just updated the OP with the rating achieved with this PR

oparoz:
Ah, I need to rescan and will post the cipher list.

oparoz:
But since all of this is going to be user-configurable, maybe I should just create another PR with just the custom logging addition, since the module needs to be compiled.

src/apache/conf/ssl.conf
```diff
@@ -88,12 +87,15 @@ SSLRandomSeed connect file:/dev/urandom 512
 
     SSLEngine on
     SSLHonorCipherOrder On
-    SSLCipherSuite ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:DH+AES:ECDH+3DES:DH+3DES:RSA+AESGCM:RSA+AES:RSA+3DES:!aNULL:!MD5:!DSS;
+	SSLCipherSuite ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA256
```
kyrofa: According to the ssllabs report in the description, this cipher suite actually decreases the cipher strength while simultaneously being incompatible with many clients. I see no upside, do you?

kyrofa: Note the the protocols are probably partially to blame there as well. We need to decide what's more important.

oparoz: It's actually the same, I still had AES128 only on my test setup when scanning.

oparoz:
Same with this which should be removed

```
    BrowserMatch "MSIE [2-5]" \
        nokeepalive ssl-unclean-shutdown \
        downgrade-1.0 force-response-1.0
```

src/apache/conf/ssl.conf
```diff
@@ -88,12 +87,15 @@ SSLRandomSeed connect file:/dev/urandom 512
...
 
     SSLCertificateFile      ${SNAP_DATA}/certs/live/cert.pem
     SSLCertificateKeyFile   ${SNAP_DATA}/certs/live/privkey.pem
     SSLCertificateChainFile ${SNAP_DATA}/certs/live/chain.pem
 
+    SSLOpenSSLConfCmd Curves brainpoolP512r1:secp521r1:brainpoolP384r1:secp384r1
```
oparoz: Apparently this doesn't work any more. Apache ignores it and offers a different list of curves...
So, the solution is to add our preferred curve at the end of the certificate which is a bit trickier, but gives us a 100% for key exchange.

oparoz: For reference, this is how it's supposed to work:
https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslopensslconfcmd

And this is the alternative that we need to use:
https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslcertificatefile

oparoz:
I did not disable session tickets, but we should do it until we rotate logs and restart Apache daily.

src/apache/conf/ssl.conf
```diff
@@ -41,7 +41,6 @@ LoadModule ssl_module modules/mod_ssl.so
 #   disabled as quickly as practical.  By the end of 2016, only the TLSv1.2
 #   protocol or later should remain in use.
 SSLProtocol all -SSLv3
-SSLProxyProtocol all -SSLv3
```
oparoz: Not needed

src/apache/conf/ssl.conf
```diff
@@ -53,7 +52,6 @@ SSLPassPhraseDialog  builtin
 #   Configure the SSL Session Cache: First the mechanism 
 #   to use and second the expiring timeout (in seconds).
 SSLSessionCache        "shmcb:${SNAP_DATA}/apache/ssl_scache(512000)"
-SSLSessionCacheTimeout  300
```
oparoz: Already the default

src/apache/conf/ssl.conf
```diff
@@ -88,7 +86,9 @@ SSLRandomSeed connect file:/dev/urandom 512
 
     SSLEngine on
     SSLHonorCipherOrder On
-    SSLCipherSuite ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:DH+AES:ECDH+3DES:DH+3DES:RSA+AESGCM:RSA+AES:RSA+3DES:!aNULL:!MD5:!DSS;
+    SSLCipherSuite ECDHE+AESGCM:DHE+AESGCM:ECDHE+AES256:DHE+AES256:ECDHE+AES128:DHE+AES:RSA+AESGCM:RSA+AES:!aNULL:!3DES:!MD5:!DSS;
```
oparoz: Remove 3DES

src/apache/conf/ssl.conf
```diff
@@ -88,7 +86,9 @@ SSLRandomSeed connect file:/dev/urandom 512
...
+    SSLOpenSSLConfCmd DHParameters "${SNAP}/conf/dhparams.2048.pem
+    SSLCompression off
```
oparoz: Disable compression server side

src/apache/conf/ssl.conf
```diff
@@ -88,7 +86,9 @@ SSLRandomSeed connect file:/dev/urandom 512
 
     SSLEngine on
     SSLHonorCipherOrder On
-    SSLCipherSuite ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:DH+AES:ECDH+3DES:DH+3DES:RSA+AESGCM:RSA+AES:RSA+3DES:!aNULL:!MD5:!DSS;
+    SSLCipherSuite ECDHE+AESGCM:DHE+AESGCM:ECDHE+AES256:DHE+AES256:ECDHE+AES128:DHE+AES:RSA+AESGCM:RSA+AES:!aNULL:!3DES:!MD5:!DSS;
+    SSLOpenSSLConfCmd DHParameters "${SNAP}/conf/dhparams.2048.pem
```
oparoz: Add 2048bits DH parameters for older clients

src/apache/conf/ssl.conf
```diff
@@ -121,43 +121,18 @@ SSLRandomSeed connect file:/dev/urandom 512
     #   o OptRenegotiate:
     #     This enables optimized SSL connection renegotiation handling when SSL
     #     directives are used in per-directory context. 
-    <FilesMatch "\.(cgi|shtml|phtml|php)$">
```
oparoz: Remove unneeded extensions

src/apache/conf/ssl.conf
```diff
@@ -121,43 +121,18 @@ SSLRandomSeed connect file:/dev/urandom 512
...
-    #     works correctly. 
-    #   Notice: Most problems of broken clients are also related to the HTTP
-    #   keep-alive facility, so you usually additionally want to disable
-    #   keep-alive for those clients, too. Use variable "nokeepalive" for this.
-    #   Similarly, one has to force some clients to use HTTP/1.0 to workaround
-    #   their broken HTTP/1.1 implementation. Use variables "downgrade-1.0" and
-    #   "force-response-1.0" for this.
-    BrowserMatch "MSIE [2-5]" \
-        nokeepalive ssl-unclean-shutdown \
-        downgrade-1.0 force-response-1.0
```
oparoz: We don't support old IE

src/apache/conf/ssl.conf
```diff
@@ -121,43 +121,18 @@ SSLRandomSeed connect file:/dev/urandom 512
...
 
     # Enable HSTS only if requested
     <IfDefine EnableHSTS>
         Header always set Strict-Transport-Security "max-age=63072000; includeSubdomains;"
     </IfDefine>
+
+	# Custom log for SSL to track ciphers used
+    CustomLog "${SNAP_DATA}/apache/logs/ssl_request_log" "%t %h %{SSL_PROTOCOL}x %{SSL_CIPHER}x \"%r\" %b"
```
oparoz: Log ciphers used.
It's really useful, but does a lot of writes and AFAIK we don't rotate logs.
Should we put it in the custom config?

src/apache/conf/ssl.conf
```diff
@@ -121,43 +121,18 @@ SSLRandomSeed connect file:/dev/urandom 512
...
+
+	# Override our settings with a user-defined configuration
+	Include ${SNAP_DATA}/apache/conf/custom_ssl.conf
```
oparoz: Override our settings with a user-defined configuration

src/apache/scripts/run-httpd
```diff
@@ -2,6 +2,16 @@
...
+# Setup SSL certificate
+RO_APACHE_CONF_FOLDER=$SNAP/conf
+RW_APACHE_CONF_FOLDER=$SNAP_DATA/apache/conf
+if [ ! -d $RW_APACHE_CONF_FOLDER ]; then
+	mkdir -p -m 750 $RW_APACHE_CONF_FOLDER
+	echo "Copying SSL config files"
+	cp $RO_APACHE_CONF_FOLDER/custom_*.conf $RW_APACHE_CONF_FOLDER
+fi
+mkdir -p -m 750 $SNAP_DATA/apache/conf
+
```
oparoz: Add the custom config to the R/W partition

src/https/utilities/https-utilities
```diff
@@ -26,6 +26,8 @@ generate_self_signed_certificate()
 	openssl req -newkey rsa:4096 -nodes -keyout $SELF_SIGNED_KEY \
 	        -x509 -days 90 -out $SELF_SIGNED_CERT -subj "/O=Nextcloud"
 
+	openssl ecparam -name secp384r1 >> $SELF_SIGNED_CERT
```
oparoz: Force the use of a stronger curve

snapcraft.yaml
```diff
@@ -92,6 +92,7 @@ parts:
       - alias
       - ssl
       - socache_shmcb
+      - log_config
```
oparoz: Needed for custom logging

src/apache/conf/httpd.conf
```diff
@@ -59,6 +59,7 @@ LoadModule dir_module modules/mod_dir.so
 LoadModule alias_module modules/mod_alias.so
 LoadModule rewrite_module modules/mod_rewrite.so
 LoadModule socache_shmcb_module modules/mod_socache_shmcb.so
+LoadModule log_config_module modules/mod_log_config.so
```
oparoz: Needed for custom logging

kyrofa:
You're covering multiple features in this PR (improving the general SSL config as well as adding logic to allow users to customize it). What happens when you open up the config more widely? Will you have multiple config files? Please try to cover one feature per PR. I'd prefer to have discussions on each, and having them both here will dilute them.

oparoz:
> You're covering multiple features in this PR (improving the general SSL config as well as adding logic to allow users to customize it).

Let's take it from a different angle then, I'm introducing SSL config customization and while I'm at it, making minor changes to the default config. I don't want to do one PR for custom logs, one for includes, one for SSL changes, mainly because the whole Snap compiling and reviewing process is a pain and we need to test the big picture. I don't see how this is any different from this massive PR which introduced a HTTPS vhost, let's encrypt, SSL config and maybe more: https://github.com/nextcloud/nextcloud-snap/pull/23

Right now, this is easy to test. You

* load it
* install a cert
* test with some browsers and sniffers
* check the SSL logs
* make a modification to the custom config
* restart Apache
* repeat the tests

Once we think this works, we can push it to the beta channel for testing and users can do the same tests.

>What happens when you open up the config more widely? Will you have multiple config files? 

Because the current setup has all the SSL changes in a special virtualhost, there has to be 2 configs. One for SSL and one for Apache in general. 
Ideally, we should remove all the global SSL elements from the virtualhost and have a single custom file. I think that would be easier for users.

oparoz:
If we agree that we'll use one custom config for Apache, then I could do a PR just for that which would need to be merged before this one if it makes it easier for you. It wouldn't work for SSL, but would just be in building block for the next steps.

oparoz:
Once #71 is merged this will be rebased.

oparoz:
Rebased against develop since we won't have custom configs for now.

src/https/utilities/https-utilities
kyrofa: [Doesn't sound like this is necessary](https://weakdh.org/sysadmin.html) as of Apache v2.4.8. `SSLOpenSSLConfCmd DHParameters` is enough. Can you verify?

src/https/utilities/https-utilities
```diff
@@ -70,6 +72,8 @@ run_certbot()
 	certbot --text --config-dir $CERTBOT_DIRECTORY/config \
 	               --work-dir $CERTBOT_DIRECTORY/work \
 	               --logs-dir $CERTBOT_DIRECTORY/logs $@
+
+	openssl ecparam -name secp384r1 >> $LIVE_CERT
```
kyrofa: Same here.

oparoz: >SSLOpenSSLConfCmd DHParameters is enough. Can you verify?

That's the theory, but it didn't work in my env. Feel free to prove me wrong :) and btw, it's another command which needs to be used for curves: `ECDHParameters`.

Also, Apache recommends doing this via adding the parameter to the cert. It may be more reliable.

oparoz:
Let's keep the conversation going regarding ECDH parameters, but let's close this. Too much unrelated stuff.

oparoz:
After having looked at what Dropbox does. It's roughly similar to what is currently in use in the snap, so let's leave it at that and update it at the next OpenSSL exploit. Advanced users will proceed differently.

PR #69, opened 2016-10-02
Add access to /media

oparoz:
Adds new storage locations capabilities to Nextcloud

* `/media` is a zone which is accessible by all snaps supporting the `removable-media` interface
* `home` is the home folder of the user and is accessible by all snaps supporting the `home` interface

Documentation on how to connect Nextcloud to these new locations is available in the wiki:
https://github.com/nextcloud/nextcloud-snap/wiki/Connecting-Nextcloud-to-the-filesystem

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -26,7 +26,7 @@ apps:
     command: start-php-fpm
     daemon: simple
     restart-condition: always
-    plugs: [network, network-bind]
+    plugs: [network, network-bind, removable-media]
 
   # redis server daemon
   redis-server:
@@ -50,7 +50,7 @@ apps:
   # Nextcloud occ command
   occ:
     command: occ
-    plugs: [network, network-bind]
+    plugs: [network, network-bind, removable-media]
 
   enable-https:
     command: enable-https


enoch85:
:+1:

oparoz:
@kyrofa - The one thing I'm unsure about is if we still need to include these plugs everywhere or not. Some of our current apps don't need to be network servers and yet they use the network-bind plug. I'm guessing we include them everywhere because they can start in random order?

kyrofa:
Access to home isn't covered by the bug. What is the purpose of that?

oparoz:
> Access to home isn't covered by the bug. What is the purpose of that?

Yes, I should add it to the issue. The goal is to give the user access to external locations and currently only "home" and `/media` are offered. "Home" is useful for people who install the Snap on their desktop.

kyrofa:
> "Home" is useful for people who install the Snap on their desktop.

I'm :-1: on this. This is a server, accessible from anywhere. With access to $HOME, the damage the snap could cause if exploited is far beyond its responsibilities. Not to mention that it would cause files in $HOME to be owned by root, and all other sorts of nonsensical things.

kyrofa:
> The one thing I'm unsure about is if we still need to include these plugs everywhere or not

Well, what will be writing files there? PHP only, I suspect? Then only PHP FPM, occ, and nextcloud-cron probably need them. But it should be thoroughly tested.

oparoz:
> Well, what will be writing files there? PHP only, I suspect? Then only PHP FPM, occ, and nextcloud-cron probably need them. But it should be thoroughly tested.

Thanks and then we'll be able to remove the network-bind permission from most apps.

> This is a server, accessible from anywhere.

The first thing to remember is that home is not automatically connected on Core, but yes, it is on Classic.
Next is that unless you connect the snap to the Internet, nobody would have access to Nextcloud, let alone the home directory.
Now if a user goes through all the above steps, he still has to connect Nextcloud to $HOME. External storage is not configured out of the box.
And finally, if it works like the `removable-media` interface, a snap is only able to access a sub-folder and not the root folder.

So, I don't see a scenario where this is a problem. I don't care much for $HOME, but I just want to make the snap flexible to use as long as we don't break it or put users at risk.

kyrofa:
> Thanks and then we'll be able to remove the network-bind permission from most apps.

That one is still required. Things still seem to use the `bind` syscall for some reason. I suspect the `network` and `network-bind` plugs will be merged at some point.

kyrofa:
> Thanks and then we'll be able to remove the network-bind permission from most apps.

That one is still required. Things still seem to use the bind syscall for some reason. I suspect the `network` and `network-bind` plugs will be merged at some point.

> The first thing to remember is that home is not automatically connected on Core, but yes, it is on Classic.

The fact that it's automatically connected anywhere is enough for me.

> Next is that unless you connect the snap to the Internet, nobody would have access to Nextcloud, let alone the home directory.

The whole point of the snap is to connect it to the internet!

> Now if a user goes through all the above steps, he still has to connect Nextcloud to $HOME. External storage is not configured out of the box.

Not true. If the interface is autoconnected, the snap has access to $HOME immediately upon install. Just because Nextcloud may not be configured to utilize it doesn't mean that $HOME is inaccessible to it.

> And finally, if it works like the removable-media interface, a snap is only able to access a sub-folder and not the root folder.

I'm not sure what you're talking about here. Have you tried it? The only thing you can't access is hidden folders within the root of $HOME. Everything else is fair game-- an attacker can wipe all music, pictures, documents, everything except SSH keys.

> So, I don't see a scenario where this is a problem. I don't care much for $HOME, but I just want to make the snap flexible to use as long as we don't break it or put users at risk.

I think this is a case of "too flexible." And I really don't see it being useful anyway.

> The folder does not need to be owned by root AFAIK, since root would be the one requesting access to it.

That's not what I said. Anything the snap _created_ in $HOME would be owned by root, so the user wouldn't have access to it.

oparoz:
Removed home, people are free to re-add it if they need it.

oparoz:
Closing since there is nothing to review

PR #71, opened 2016-10-02
Make it possible to override the read-only Apache configuration

### LINKED ISSUE #70: Split Apache config in two to make it possible for users to customise it
* `httpd.conf` will have an extra include which will load the user's `custom.conf`.
* Some SSL variables will have to be moved outside the SSL virtualhost in order to be "overridable".
* Offer some example of customisation

oparoz:
This will be required when auto-tuning configurations as we won't always get it right.

* Includes a safer, custom SSL config which can be enabled by the user
* Includes 2048 and 4096 bits DH params, ready to be used
* Includes SSL ciphers logging example to help with debugging
* Removed configuration errors and some variables for which values were already the default

Fixes #70 

Tested on Chrome, Firefox, Edge, IE11

**[Documentation on how to customize the SSL config as an example](https://github.com/nextcloud/nextcloud-snap/wiki/How-to-modify-your-SSL-configuration)**

## Safer SSL config example

secure_config

People who have banned OSX clients and don't require compatibility with Android < 4.4 can tighten their SSL config by uncommenting some proposed changes which are included in the custom.conf file that the user can edit.

### Changes
* TLSv1.2 only
* ECDH only
* Drop support for HMAC-SHA1
* Disable session ticket
* Enable OCSP stapling

```
# Warning the Mac OSX client currently requires TLSv1
#SSLProtocol All -SSLv3 -TLSv1 -TLSv1.1
#SSLCipherSuite ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA256
# Do not uncomment if only using the recommended ECDHE algorithm for key exchange, as shown above in SSLCipherSuite
# We provide both dhparams.2048.pem and dhparams.4096.pem parameters
#SSLOpenSSLConfCmd DHParameters "${SNAP}/conf/dhparams.4096.pem"
# Using SSL Session Tickets without restarting the web server with an appropriate frequency (e.g. daily) compromises perfect forward secrecy.
# Comes with a performance penalty.
# Should not be needed once https://github.com/nextcloud/nextcloud-snap/issues/65 is implemented
#SSLSessionTickets off

# OCSP stapling. See https://en.wikipedia.org/wiki/OCSP_stapling
# WARNING: Only enable with a real certificate, like the one provide by Let's Encrypt
#SSLUseStapling on
#SSLStaplingResponderTimeout 5
#SSLStaplingReturnResponderErrors off
#SSLStaplingCache "shmcb:${SNAP_DATA}/apache/stapling_cache(128000)"

# Custom log for SSL to track ciphers used
#CustomLog "${SNAP_DATA}/apache/logs/ssl_request_log" "%t %h %{SSL_PROTOCOL}x %{SSL_CIPHER}x \"%r\" %b"
```

ssllabs_a


src/apache/conf/custom.conf
```diff
@@ -0,0 +1,42 @@
...
+# alert, emerg.
+#
+# LogLevel warn
+
+#########################
+#          SSL          #
+#########################
+# Warning the Mac OSX client currently requires TLSv1
+#SSLProtocol All -SSLv3 -TLSv1 -TLSv1.1
+#SSLCipherSuite ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA256
```
kyrofa: Adding such things to this file is a bad idea. Remember you can't update it after it's installed, and I guarantee our recommended cipher suite will change over time. I suggest you simply put a link to the wiki with some sort of "see the following link for example configurations," and put this information on a wiki you can keep up-to-date.

src/apache/scripts/run-httpd
```diff
@@ -2,6 +2,16 @@
 
 mkdir -p -m 750 $SNAP_DATA/apache/logs
 
+# Setup SSL certificate
```
kyrofa: This doesn't really have anything to do with the SSL certs, does it?

src/apache/scripts/run-httpd
```diff
@@ -2,6 +2,16 @@
...
+RO_APACHE_CONF_FOLDER=$SNAP/conf
+RW_APACHE_CONF_FOLDER=$SNAP_DATA/apache/conf
+if [ ! -d $RW_APACHE_CONF_FOLDER ]; then
+	mkdir -p -m 750 $RW_APACHE_CONF_FOLDER
+	echo "Copying SSL config files"
```
kyrofa: Ditto above, nothing to do with SSL.

src/apache/scripts/run-httpd
```diff
@@ -2,6 +2,16 @@
...
+	cp $RO_APACHE_CONF_FOLDER/custom.conf $RW_APACHE_CONF_FOLDER
+fi
+mkdir -p -m 750 $SNAP_DATA/apache/conf
```
kyrofa: This line is likely unnecessary given line 9.

src/apache/scripts/run-httpd
oparoz: :+1:

src/apache/scripts/run-httpd
oparoz: :+1:

src/apache/conf/custom.conf
oparoz: I'm torn on this, because people plugging in a monitor and a keyboard and willing to modify their configs will find it easier to just uncomment what's there.
And I don't think the config is going change that often. Maybe twice a year and I suspect that people changing it will be people who keep track of these things.

What I'd like to introduce at some point is a way to push notifications from the Snap to Nextcloud. That way we can tell people about changes which have an impact on their custom configs, like if a major vulnerability in OpenSSL is published.

oparoz:
After having discussed it with @kyrofa, it was decided that this would not be part of this snap.

PR #72, opened 2016-10-02
Spreedme integration

### LINKED ISSUE #64: Add spreed.ME reverse proxy to Apache
We need to connect to the Spreed.ME Snap through a common proxy, so it's best to add it to the one used by Nextcloud. The configuration is available here:
https://github.com/strukturag/nextcloud-spreedme/blob/master/doc/example-config-apache.md
- [x] Add 2 new Apache modules
- [x] Provide configuration

This is to be added either to the default or to the custom config. There shouldn't be any harm in having a location which leads to nowhere.

```
# Provide access to Spreed.ME
<Location /webrtc>
    ProxyPass http://127.0.0.1:8080/webrtc
    ProxyPassReverse /webrtc
</Location>

<Location /webrtc/ws>
    ProxyPass ws://127.0.0.1:8080/webrtc/ws
</Location>

ProxyVia On
ProxyPreserveHost On
RequestHeader set X-Forwarded-Proto 'https' env=HTTPS
```

oparoz:
Fixes #64 

The proxy configuration is commented out in the Apache `custom.conf` since users will need to use the command line to integrate Nextcloud with Spreed.ME anyway.

I don't feel comfortable enabling a connection to localhost:8080 by default. We don't know what might end up being loaded here via another Snap.

Documentation and test snaps are available here: https://help.nextcloud.com/t/spreed-me-snap-for-the-nextcloud-snap-beta/3757

We will re-evaluate this for Nextcloud 11.

## Todo

- [ ] Will be rebased once #71 is merged
- [ ] Update documentation to let people know they need to uncomment the Apache config

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -82,6 +82,8 @@ parts:
       - headers
       - proxy
       - proxy_fcgi
+      - proxy_http
+      - proxy_wstunnel
       - setenvif
       - env
       - rewrite
--- a/src/apache/conf/httpd.conf
+++ b/src/apache/conf/httpd.conf
@@ -54,6 +54,8 @@ LoadModule headers_module modules/mod_headers.so
 LoadModule setenvif_module modules/mod_setenvif.so
 LoadModule proxy_module modules/mod_proxy.so
 LoadModule proxy_fcgi_module modules/mod_proxy_fcgi.so
+LoadModule proxy_http_module modules/mod_proxy_http.so
+LoadModule proxy_wstunnel_module modules/mod_proxy_wstunnel.so
 LoadModule unixd_module modules/mod_unixd.so
 LoadModule dir_module modules/mod_dir.so
 LoadModule alias_module modules/mod_alias.so
--- a/src/apache/conf/ssl.conf
+++ b/src/apache/conf/ssl.conf
@@ -160,4 +160,20 @@ SSLRandomSeed connect file:/dev/urandom 512
     <IfDefine EnableHSTS>
         Header always set Strict-Transport-Security "max-age=63072000; includeSubdomains;"
     </IfDefine>
+
+	#########################
+	#       Spreed.ME       #
+	#########################
+	<Location /webrtc>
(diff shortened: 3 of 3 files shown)


kyrofa:
@oparoz can you please fork and propose from there? Putting your branches directly in the project clutters it up and wreaks havoc with our CI that replicates pushes to LP.

oparoz:
I don't understand what the problem is. The fact that new builds are generated for each branch is actually the wanted effect, no? It also makes it easier to collaborate since other community members can push to those branches. Travis seems fine, so is it Launchpad having issues?

oparoz:
I've just looked at the Travis history and apparently you never build test snaps automatically, so please explain to me what the problem is with CI, because I think you should create branches on here as well :D

kyrofa:
All the new branches [here](https://code.launchpad.net/nextcloud-snap) is the issue. We have no way to cull them, so it's easier to just use the project for branches that are setup for automatic builds.

oparoz:
Is it just a visual thing? On the server side we currently have 57 branches, but it's quick to find what we need on Github and collaborators can quickly pull down a project branch. I don't expect anybody to use Launchpad to get the source code since it's not where it's being developed.
The whole point of having CI right now is to push branches to LP so that they're automatically built so that we can test new features without any manual intervention.
Maybe we could use some pre-determined prefixes to clearly identify PRs? I still see only drawbacks in using forks for PRs.

oparoz:
This has been modified to not use custom configs any more and we cannot use a unix socket to serve the required HTML content.

kyrofa:
I don't like shipping useless configs by default (i.e. this PR does nothing unless spreed is installed). As you mention in the description, it widens the attack surface since we don't actually know what's there. If we want to enable this, I suggest we tie it into the [`configure` hook](https://github.com/snapcore/snapd/blob/master/docs/hooks.md#configure) so that users can enable it if they want it.

Also, since the PR description and conversation is a bit unrelated to the current state of the PR, combined with the fact that we're using a forking flow, I'm going to go ahead and close this while you rework. Please re-propose from a fork.

PR #76, opened 2016-10-03
Upgrade to Nextcloud 10.

### LINKED ISSUE #40: Update to Nextcloud 10

kyrofa:
This PR resolves #40 by upgrading to Nextcloud 10.0.1, which means we can finally use vanilla upstream and enable the integrity check!

Test it out on amd64 with [this snap](http://people.canonical.com/~kyrofa/nextcloud-box/pr_76.snap).

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -119,11 +119,11 @@ parts:
 
   nextcloud:
     plugin: copy
-    source: https://github.com/kyrofa/nextcloud.git
-    source-tag: 9.0.53
+    source: https://download.nextcloud.com/server/releases/nextcloud-10.0.1.tar.bz2
     files:
       '*': htdocs/
       '.htaccess': htdocs/
+      '.user.ini': htdocs/
 
   php:
     plugin: php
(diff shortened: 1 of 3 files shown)


snapcraft.yaml
```diff
@@ -119,11 +119,11 @@ parts:
 
   nextcloud:
     plugin: copy
-    source: https://github.com/kyrofa/nextcloud.git
-    source-tag: 9.0.53
+    source: https://download.nextcloud.com/server/releases/nextcloud-10.0.1.tar.bz2
     files:
       '*': htdocs/
       '.htaccess': htdocs/
+      '.user.ini': htdocs/
```
oparoz: It's a bad idea to import this file since it's read-only and will override what we define in the PHP config.

kyrofa: If we leave it out the integrity check fails. What's more important?

oparoz: OK. We can have both. Set `user_ini.filename` to `""` and the file won't be loaded :)

kyrofa: Good call, that seems to work. I've updated the PR, as well as the reference snap.

oparoz: Great :)

PR #82, opened 2016-10-06
Update PHP to 7.0.11

### LINKED ISSUE #57: Update PHP to the latest version
The currently shipped 7.0.2 contains several security vulnerabilities.
There was a conflict previously with the local domain used, but this needs to be re-tested.

oparoz:
Fixes #57


kyrofa:
:+1: looks like the problem with the earlier version was solved!

PR #83, opened 2016-10-07
Add proxy for snapweb

### LINKED ISSUE #79: Add proxy to the snapweb app store
Just like for Spreed.ME, we have to add a proxy to the snapweb snap, but this time it's mandatory.

oparoz:
This makes it possible to integrate the snap store inside Nextcloud

Fixes #79

### DIFF
--- a/src/apache/conf/ssl.conf
+++ b/src/apache/conf/ssl.conf
@@ -161,3 +161,17 @@ SSLRandomSeed connect file:/dev/urandom 512
         Header always set Strict-Transport-Security "max-age=63072000; includeSubdomains;"
     </IfDefine>
 </VirtualHost>
+
+#########################
+#        Snapweb        #
+#########################
+Listen 4201
+<VirtualHost *:4201>
+    SSLEngine on
+    SSLCertificateFile      ${SNAP_DATA}/certs/live/cert.pem
+    SSLCertificateKeyFile   ${SNAP_DATA}/certs/live/privkey.pem
+    SSLCertificateChainFile ${SNAP_DATA}/certs/live/chain.pem
+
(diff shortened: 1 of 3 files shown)


src/apache/conf/ssl.conf
```diff
@@ -161,3 +161,17 @@ SSLRandomSeed connect file:/dev/urandom 512
         Header always set Strict-Transport-Security "max-age=63072000; includeSubdomains;"
     </IfDefine>
 </VirtualHost>
+
+#########################
+#        Snapweb        #
+#########################
+Listen 4201
+<VirtualHost *:4201>
```
kyrofa: Obviously snapweb is already listening (with no SSL) on 4200. What is the purpose of also having it here? Just to add SSL to it?

oparoz: Without SSL support, we cannot integrate it inside Nextcloud, so we need to create a virtualhost on our reverse proxy.

kyrofa: This also means one can only use snapweb from within nextcloud if SSL is enabled, right? Isn't that a little odd?

oparoz: No, you have to use 2 different URLs, one for SSL and one for without SSL. IIRC, only the port changes.

kyrofa: Note that Snapweb has gained HTTPS. This probably needs to be re-evaluated.

oparoz: Yep, I'd say it's not needed any more for most users if they can simply add the URL. Enterprise users might need to create their own vhost, but it's out of sope for this project.

PR #90, opened 2016-10-09
Add with-libdir to PHP plugin

### LINKED ISSUE #88: Add with-libdir to PHP plugin
`--with-libdir` is required for the compilation of some PHP features. Since it's architecture dependant, it should be added to the plugn itself.

oparoz:
Fixes #88

### DIFF
--- a/parts/plugins/x-php.py
+++ b/parts/plugins/x-php.py
@@ -59,6 +59,7 @@ class PhpPlugin(autotools.AutotoolsPlugin):
         return schema
 
     def __init__(self, name, options, project):
+        options.configflags.extend(['--with-libdir=lib/{}'.format(project.arch_triplet)])
         super().__init__(name, options, project)
 
         self.extensions_directory = os.path.join(self.partdir, 'extensions')


oparoz:
I'm sure it could be improved by checking if it hasn't already been set, but it's not a problem to have it twice and developers will only add it if compilation of a new option fails.

parts/plugins/x-php.py
```diff
@@ -59,6 +59,7 @@ def schema(cls):
         return schema
 
     def __init__(self, name, options, project):
+        options.configflags.extend(['--with-libdir=lib/{}'.format(project.arch_triplet)])
```
kyrofa: I don't think this will be correct on 64-bit systems. From [here](https://secure.php.net/manual/en/configure.about.php):

> Specifies the directory where the libraries to build PHP exists on a Unix system. For 64bit systems, its needed to specify this argument to the lib64 directory like: --with-libdir=lib64.

oparoz: From my understanding "lib64" was the old way of doing things. I've tested this solution on amd64 and it worked. Haven't tried on ARM yet.

oparoz: Just checked 2 systems and none of them have a lib64 or lib32 folder in `/usr` , but both have the `/usr/lib/<arch_triplet>` folder

kyrofa: Huh, okay. Another question: this is adding `--with-libdir` to the build of PHP itself, when the examples I see using it are on the build of individual PHP extensions. Is this flag in the wrong spot?

oparoz: Good point. You need this when enabling some features which ship with PHP, but it might also be needed for extensions.
What the best way to do this? Simply duplicate the line?
Thinking about lib64, I think it didn't work and that's the reason I had to try something else, but before I try again. How would we get the proper path based on the arch since lib32 doesn't exist on armhf?

kyrofa: Yeah, I'd just add it to where the extensions are being built.

Regarding the path, you can interrogate the path itself via python, or use an if/else based on the arch. But I'd make sure the docs are correct before going to that effort.

kyrofa:
Closing due to inactivity.

PR #110, opened 2016-10-19
Add ability to access removable media.

### LINKED ISSUE #66: Add access to /media
`/media` is a zone which is accessible by all snaps supporting the `removable-media` interface. We should add it to this Snap.
## Benefits
- Creates a permanent storage location. Currently, when adding other Nextcloud snaps, they're all empty
- Provides an area which will survive a snap removal, whether by accident or because it has become unrecoverable
- Allows Nextcloud to have access to data of other snaps
- Gives access to mounted storage
- Makes it possible to create backups
- Gives access to USB sticks?

kyrofa:
Resolve #66 by adding the `removable-media` plug to the required apps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,13 @@ there's no way for the snap to change the setting from Ajax to Cron for you.
 You must do that manually in the admin interface if you want to take advantage
 of the performance improvements.
 
+Also note that the interface providing the ability to access removable media is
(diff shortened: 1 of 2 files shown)


oparoz:
:+1: Works

kyrofa:
Thanks @oparoz!

PR #146, opened 2016-12-13
Update to the new python plugin

elopio:


elopio:
This change requires snapcraft 2.17 or greater.

PR #148, opened 2016-12-15
Added mysqldump command

### LINKED ISSUE #126: Add mysqldump command
In the documentation it suggests mysqldump as the way to backup the database:
https://docs.nextcloud.com/server/9/admin_manual/maintenance/backup.html

However, this snap only exposes the mysql command. I'm not sure if it is possible to backup a mysql database using just the mysql command, but if not then the mysqldump command would be needed. Knowing nothing about mysql I don't know the best way to backup the database, but it strikes me as an important thing to do! I'm guessing that if mysqldump is installed as part of mysql it is just a case of exposing it in the YAML file, and I'd be happy to have a go at that myself...

thomascobb:
Fixes #126

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -49,6 +49,10 @@ apps:
     command: mysql --defaults-file=$SNAP_DATA/mysql/root.ini
     plugs: [network, network-bind]
 
+  mysqldump:
+    command: mysqldump --defaults-file=$SNAP_DATA/mysql/root.ini --lock-tables nextcloud
+    plugs: [network, network-bind]
+
   # Nextcloud occ command
   occ:
     command: occ
@@ -255,7 +259,6 @@ parts:
       - -bin/mysql_client_test
       - -bin/mysql_config*
       - -bin/mysqld_multi
-      - -bin/mysqldump*
       - -bin/mysqlimport
       - -bin/mysql_install_db
       - -bin/mysql_plugin


thomascobb:
This currently spits out the database dump to stdout. Is this what we want? You would have to run it like:

``sudo nextcloud.mysqldump > nextcloud-sqlbkp_`date +"%Y%m%d"`.bak``

thomascobb:
It also expects the database name to be called nextcloud as that's what the yaml file seems to create

kyrofa:
> This currently spits out the database dump to stdout. Is this what we want?

*shrug* that's how `mysqldump` works, I don't have a problem with it.

However, I've been thinking about this. What is the goal here? To backup/migrate a snap install to another snap install? Or migrate to another install that isn't snap-based? Do we care about the database specifically, or do we want _everything_ that is Nextcloud?

I ask this because we could make this a lot more friendly-- we could package up the nextcloud config and database together into one backup tarball, and write an importer as well. We could do this for the raw data as well, though that tarball could be huge.

What are your thoughts on that?

kyrofa:
Aw, you pretty much asked that in the bug and I missed it, my apologies, I was on vacation for a while :frowning_face: .

kyrofa:
I'm going to go ahead and merge this; it'll be handy to have until we have a read importer/exporter.

Issue #172, opened 2017-01-08
OCR Integration (dependencies outside Snap container?)


adamlove86:
Hi there,

Got Nextant working well, and recently upgraded to Nextcloud Snap 11. I've also installed OCR version 2, however dependencies required for OCR are:
- OCRmyPDF
- tesseract-ocr

See (https://github.com/janis91/ocr)

Whilst I've installed these dependencies, they won't connect to the nextcloud snap, or I'm not sure how to make them talk to each other (or how to install them inside the snap container).

Any assistance would be appreciated!

kyrofa:
Yeah, those things should probably be contained within their own snap. Unfortunately, snapd provides no way to get the app contained within said new snap to get into nextcloud. The problem is difficult to solve in a secure manner without that, and it's a known issue, so let's wait for a fix.

kyrofa:
Tracking upstream bug: LP: [#1655125](https://bugs.launchpad.net/snappy/+bug/1655125)

adamlove86:
Came across these snaps the other day - you may already be across it though as he also works at Canonical

https://uappexplorer.com/app/nextcloud-nextant.rmescandon

https://uappexplorer.com/app/solr.rmescandon

kyrofa:
Yep, he ran into the same issue.

adamlove86:
Thanks for your help with this. 
I've moved on from the Snap to the full Nextcloud and all seems to be okay. Feel free to close if you wish as it's no longer an issue for me and there hasn't been any feedback from anyone else regarding this issue.

PR #182, opened 2017-01-20
Search support using Solr + nextant

rmescandon:
Features:
- document search and inner document word search
- suggested words while typing

Uses:
- Solr server
- Nextant app

New files:
- src/nextant/scripts/nextant-deploy (nextant-deploy daemon)
- src/solr/scripts/custom-solr (solr start/stop script based in the original one included in solr. This overrides the default for Solr to be started up in a snap)
- src/solr/scripts/solr.in.sh (overrides default Solr configuration with writable paths when needed)
- src/solr/scripts/sorl-server (launcher for Solr. will call 'custom-solr' as needed to start, stop or create solr core)

Daemons:
- Solr daemon of type forking and always up, needed for holding indexed documents. First time, it creates a core for nextant. Rest of times starting up, reuses that same core
- Nextant-deploy, executed only when snap is installed, to deploy and configure nextant and link it to Solr. Steps in that process are:
1) Wait for nextcloud be configured and have available extra-apps folder
2) Copies nextant code to extra-apps
3) Tries to enable nextant. If cannot, it will be retrying every 30 seconds (app cannot be enabled until nextcloud is completely configured by UI)
4) Configures nextant with default values
5) Indexes current nextcloud documents (Documents in nextcloud will be indexed every certain time. This can be configured in UI settings, later)
6) modifies some solr config files to enable suggested words when searching
7) restart Solr server


kyrofa:
@rmescandon you've done an amazing job of figuring out the snap and integrating several new components into it, and I'm sure that took a lot of work. However, I wish you would have logged a bug before going to all this effort, because I would have marked it as a dupe of #172.

Long story short: adding/enabling every Nextcloud app under the son within the snap wouldn't scale-- it would soon be huge, filled with things most people don't want. Heck, not even all the default apps included with Nextcloud are completely supported (see #117).

As discussed in #172, once snapd has a solution to LP: [#/1655125](https://bugs.launchpad.net/snappy/+bug/1655125), you'll be able to package Solr and Nextant in their _own_ snap and plug into Nextcloud. Until then, sit tight, and hold onto this code, but it's not going to work for this snap. I'm sorry for the trouble, and I appreciate the work put into this.

PR #201, opened 2017-01-27
https: support custom certificates.

### LINKED ISSUE #199: Using your own SSL certs in snap
How can a person use their own SSL cert in the nextcloud snap? I had generated self-signed thinking I could juts replace them - but no go. (read only Filesystem). We have a corporate wildcard we want to use.

kyrofa:
This PR fixes #199 by adding an option to the `nextcloud.enable-https` script to use custom certificates. This actually involved a bit of a rewrite of the `nextcloud.enable-https` script. Instead of using purely opaque options, it now uses subcommands with their own options and help output. This hopefully makes it easier to use. One can now use it with custom certificates like so:

```bash
$ sudo nextcloud.enable-https custom cert.pem key.pem chain.pem
```

Make the chain.pem just the cert.pem if you don't have a chain from your CA. If the certificate isn't self-signed and you'd like to enable HSTS, you can do so with the `-s` option:

```bash
$ sudo nextcloud.enable-https custom -s cert.pem key.pem chain.pem
```

Note that these files must be in a location that is readable by the snap (the help text should give you some pointers, `sudo nextcloud.enable-https custom -h`).


npodbielski:
Does it still works with current version of nextcloud?
I am using 15.0.4

kyrofa:
@npodbielski it certainly should, anyway. Are you having troubles? Please log an issue if so.

npodbielski:
I had problems because in docs or in help there is not mention that certificate files should be named exactly the same as `cert.pem key.pem chain.pem` and already be placed in /nextcloud directory. This is strange and surprising that you can not use some arbitrary location. Or maybe this is a bug since this works till the point when apache is starting and cannot find i.e. `cert.pem` because command got some other name of file. 
So it works with default name but not any other.

kyrofa:
> I had problems because in docs or in help there is not mention that certificate files should be named exactly the same as `cert.pem key.pem chain.pem`

That's because there is [no such requirement](https://github.com/nextcloud/nextcloud-snap/pull/201/files#diff-98f2127fed8e6af26801dea6cac1fa6aR66).

> and already be placed in /nextcloud directory

THAT is the requirement. Well, almost. Read the help text:

```
$ sudo nextcloud.enable-https custom -h
Usage:
    nextcloud.enable-https custom [-h -s] <cert> <key> <chain>

    Use certificates generated by other means. Note that
    the files provided to this command must be readable
    by the snap, which means they must contained in one
    of four directory trees:
      - /var/snap/nextcloud/current
      - /var/snap/nextcloud/common
      - /root/snap/nextcloud/11688
      - /root/snap/nextcloud/common

    Also note that this command will create copies of the
    files provided; if this command completes
    successfully, they can be safely removed.

    -h: Display this help message.
    -s: Enable HTTP Strict Transport Security (HSTS)
        (default is off-- leave off if self-signed).
```
The certs must be readable by the snap in order for you to successfully run this command, which means the certs need to be placed into one of those four locations. The snap, being confined, can't read anything else.

npodbielski:
I head them in media, but probably before I allowed nextcloud to read it. Anyway I got confusing errors:
obraz
So it was able read and check certificate and it failed miserably. :)

NeghVar:
Here's what I do not understand.  There is a syntax error in the ssl.conf file.  But the file is read-only.  I could not even override in root using :wq! in vi.  The error I get is that it is in a read-only file system.  Checked permissions.  700.  So it sounds like a catch-22.  ssl.conf has a syntax error that needs correcting, yet the file is locked down, thus cannot be corrected.  
BTW, cert.pem is present in the folder, and it has the same key as my original crt file (<domain>.crt)

kyrofa:
@NeghVar it's not really a syntax error: in @npodbielski's case, the cert wasn't properly put in place and so Apache can't find it. Why it calls that a syntax error, I don't know. What _your_ situation is, I don't know. May I suggest opening an issue so we can chat about it?

Integrion:
Hi, this config feature is either broken, or perhaps needs clearer explanation. I attempted to do a custom cert install using non-self-signed certificates.  I manually copied the required certificate files (the cert, chain and privkey files) into one of the 4 directories listed, as indicated when running the -h option. Our files however were not named cert.pem, chain.pem and privkey.pem. I presumed that they could be named anything we want, as long as they were provided on the command line in the proper order:  ie: <cert> <key> <chain>  This however did not work.  To finally get things to work,  we first had to copy our cert files to '/var/snap/nextcloud/12387/certs/live/, and then we had to rename them respectively to cert.pem, privkey.pem and chain.pem. Once we did that, our custom certs finally worked.  I noticed that when I attempted to manually edit the ssl.conf file, even when logged in as root, that the file appeared to be locked by some other running process. I'm thinking that perhaps some kind of issue exists with the nextcloud.enable-https script from copying and writing the required config changes, hence why I needed to manually copy and rename the certs files into the correct final directories pointed to by the ssl.conf file. Thanks.

npodbielski:
@kyrofa I recently tried to automate certificate update and everythings seems to works fine, except when I open nextcloud page in the browser I still have old certificate. I am doing something like this for an update:
```
cp -f /media/cert/nextcloud/nextcloud_cert.pem /snap/nextcloud/cert/cert.pem
cp -f /media/cert/nextcloud/nextcloud_chain.pem /snap/nextcloud/cert/chain.pem
cp -f /media/cert/nextcloud/nextcloud_privkey.pem /snap/nextcloud/cert/key.pem
nextcloud.enable-https custom -s /snap/nextcloud/cert/cert.pem /snap/nextcloud/cert/key.pem /snap/nextcloud/cert/chain.pem 
sudo snap restart nextcloud
```
This results in something like this:
```
Installing custom certificate... done
Restarting apache... done
Restarted.
```
after that in browser page still uses old certificate. Can I check somehow in snap directory if correct certificate is in use? 
Is this possible that Apache is using old certificate which is still valid for few days and will switch to new one after it will no longer be?

kyrofa:
@npodbielski `/snap/nextcloud/cert/` isn't a location the snap can read. Instead of using that, try using e.g. `/var/snap/nextcloud/current/cert`, maybe.

npodbielski:
@kyrofa ok thanks for info. I figure it out. In my case your script `nextcloud.enable-https custom -s` does not copy certificate files at all. It does not matter where I put them. i.e. i tried to put files into
`/media/` - snap have access there because i have all my data in this directory tree - no luck
`/var/snap/nextcloud/current/certs/new` - none of files were replaced
`/var/snap/nextcloud/current/certs/custom` - when files were named diffrent (i.e. `key.pem` instead of `privkey.pem`) it did not work and apache did not start at all
But when I just replace the files manually, with names exactly as old ones, it works after restart with no problem without call to `nextcloud.enable-https` even. 
So `nextcloud.enable-https` just to configure nextcloud and certificate files has to be *exactly* in right directory and *exactly* with right names.
Probably had the same problem few months back.

vcollerji:
@kyrofa, thank you so much for your work and patient descriptions. I just used it to bring up a Nextcloud instance (via snap) on Ubuntu 20.04. Everything worked like a dream! Thanks again!

PR #207, opened 2017-02-13
exposed nextcloud documents through content interface

rmescandon:
Sharing documents through content interface.
Needed for collabora office snap (code) to communicate to nextcloud

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -6,6 +6,13 @@ description: |
  more at home and in your enterprise.
 confinement: strict
 
+slots:
+  nextcloud-data:
+    content: nextcloud-documents
+    interface: content
+    write:
+      - $SNAP_COMMON/nextcloud/data
+
 apps:
   # Apache daemon
   apache:


kyrofa:
Please read the [contributing guide](https://github.com/nextcloud/nextcloud-snap/blob/master/CONTRIBUTING.md). I prefer to have discussions on bugs before you go to the work of making a PR. I think [I mentioned this on your last PR](https://github.com/nextcloud/nextcloud-snap/pull/182#issuecomment-274229392).

rmescandon:
Sorry @kyrofa, you're completely right. I forgot to file a related bug. 
Fixes [#216] [1]

[1] https://github.com/nextcloud/nextcloud-snap/issues/216

kyrofa:
> This should not be merged. We should never give access to the data folder to outside apps.

Note that this interface is not connected unless the user explicitly does so. But yes, I'm not a fan of this either. Since there's conflicts here anyway I'll go ahead and close this. We can continue discussions on the bug as necessary.

oparoz:
> Note that this interface is not connected unless the user explicitly does so.

Yes, I should have been clearer. The problem is more that it breaks Nextcloud.

PR #251, opened 2017-04-27
apache: disable HTTP methods TRACE and TRACK

### LINKED ISSUE #250: Disable TRACE and TRACK
From the OpenVAS report:

> Debugging functions are enabled on the remote HTTP server.
> 
> The remote webserver supports the TRACE and/or TRACK methods. TRACE and TRACK are HTTP methods which are used to debug web server connections.
> 
> It has been shown that servers supporting this method are subject to cross-site-scripting attacks, dubbed XST for Cross-Site-Tracing, when used in conjunction with various weaknesses in browsers.
> 
> An attacker may use this flaw to trick your legitimate web users to give him their credentials.

Verify TRACE/TRACK status with this script (run both with and without HTTPS enabled) (depends upon curl and grep):

```bash
#!/bin/sh

if [ $# -ne 1 ]; then
        echo "You must provide a domain or IP address" >&2
        exit 1
fi

if $(curl -k -s -I -L -X TRACE $1 | grep -q "HTTP.*200"); then
        echo "TRACE is not disabled!"
else
        echo "TRACE is disabled: good."
fi

if $(curl -k -s -I -L -X TRACK $1 | grep -q "HTTP.*200"); then
        echo "TRACK is not disabled!"
else
        echo "TRACK is disabled: good."
fi
```

kyrofa:
TRACE and TRACK are HTTP debugging methods, and it has been shown that they can be used for cross-site-scripting attacks. These methods are not necessary for a properly-functioning Nextcloud install, so this PR resolves #250 by disabling them.

### DIFF
--- a/src/apache/conf/httpd.conf
+++ b/src/apache/conf/httpd.conf
@@ -180,6 +180,14 @@ LogLevel warn
 #
 TypesConfig conf/mime.types
 
+# Disable HTTP TRACE method.
+TraceEnable off
+
+# Disable HTTP TRACK method.
+RewriteEngine On
+RewriteCond %{REQUEST_METHOD} ^TRACK
+RewriteRule .* - [R=405,L]
+
 # Only enable SSL if requested
 <IfDefine EnableHTTPS>
     Include ${SNAP}/conf/ssl.conf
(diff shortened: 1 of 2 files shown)


kyrofa:
Note that rewrite inheritance didn't seem to work for me here, so I ended up just copying the TRACK-disabling rewrite rule to each vhost. Annoying, but at least it's a small rule.

Issue #329, opened 2017-07-30
Use the upstream repository for mysql


pachulo:
I think that we should use the upstream repository of [mysql-server](https://github.com/mysql/mysql-server) and then apply the needed patch at compile time to make it work https://github.com/kyrofa/mysql-server/commit/bb6c86ca997b2ca1b052cb83e91152220fe149ad in the snap sandbox.

This should make mysql updates less painful.

PR #330, opened 2017-07-30
Use a patch for mysql-server instead of kyrofa's custom repository

pachulo:
This PR changes the repository from where mysql-server is pulled and adds kyrofa's patch to make it run in heavily confined systems like Ubuntu Core.

Solves #329

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -235,10 +235,11 @@ parts:
 
   mysql:
     plugin: cmake
-    source: https://github.com/kyrofa/mysql-server.git
+    source: https://github.com/mysql/mysql-server
     source-type: git
-    source-branch: mysql-5.7.18_support_no_setpriority
-    after: [boost]
+    source-tag: mysql-5.7.18
+    prepare: git apply $SNAPCRAFT_STAGE/support-compile-time-disabling-of-setpriority.patch
+    after: [boost, patches]
     configflags:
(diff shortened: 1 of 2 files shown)


snap/snapcraft.yaml
```diff
@@ -235,10 +235,11 @@ parts:
 
   mysql:
     plugin: cmake
-    source: https://github.com/kyrofa/mysql-server.git
+    source: https://github.com/mysql/mysql-server
     source-type: git
-    source-branch: mysql-5.7.18_support_no_setpriority
-    after: [boost]
+    source-tag: mysql-5.7.18
+    prepare: git apply $SNAPCRAFT_STAGE/support-compile-time-disabling-of-setpriority.patch
```
kyrofa: I'm afraid you've hit the limits of patching in `prepare` here: it doesn't work for all plugins, most notably those that do out-of-source builds, like cmake. You can see this in the CI results: you're not building a patching mysql. Some context can be found [in this PR](https://github.com/snapcore/snapcraft/pull/1414), but long story short, this isn't possible without some hackery for now. There is a `post-pull` [in the pipeline](https://forum.snapcraft.io/t/scripting-and-setting-metadata-in-snapcraft/1447), I suggest we hold off on this patching until we can use that new functionality.

pachulo:
OK, good to know about this!

PR #336, opened 2017-08-11
Update snapcraft.yaml

popey:
Update the description for the snap, which is more appealing when discovered via "snap list" or "snap find".

Note that due to a store limitation, the text will need to be directly updated in the store dashboard, as changes to the yaml do not override the store.


kyrofa:
Thanks @popey, looks good to me, although this PR need to be to develop instead of master. Mind doing that?

PR #349, opened 2017-09-11
apache: support changing ports

### LINKED ISSUE #134: Change apache listen port
Hi.

Install Nextcloud on a raspberry pi3, with a NexcloudBox, I made the installation from zeros using a preinstalled version of ubuntu 16.04.

I have made everything work but for external access I need to do a port redirection on the router of my ISP provider. I can not leave the port 80 configured in apache since that port is used by the provider. When trying to change the apache configuration in the httpd.conf file located in /snap/nextcloud/173/conf/httpd.conf is in a read-only folder.

How can I do to change the listening port of apache?

Thank you.

kyrofa:
Currently, Apache listens on port 80. If HTTPS is enabled, it listens on both 80 and 443. This is inflexible, especially if one wants to put the snap behind a proxy of some kind.

This PR fixes #134 by adding support for changing HTTP and HTTPS ports via the `configure` hook. It exposes these configurations via the `ports.http` and `ports.https` options, respectively.

Further testing is necessarily before this can land. Please test using the `stable/pr-349` channel:

    $ sudo snap install nextcloud --channel=stable/pr-349

    $ sudo snap refresh nextcloud --channel=stable/pr-349

See the README for details on how to use.


kyrofa:
Alright, description updating with testing details.

extraymond:
Hi! I tried to test this by refreshing my nextcloud to the  one given above.

And I receive the following error:

```
error: cannot perform the following tasks:
- Run configure hook of "nextcloud" snap (snap "nextcloud" has no "configure" hook)
```

kyrofa:
@extraymond thanks for trying it! Can you please tell me:

1. What architecture are you on?
2. What is the revision of the Nextcloud snap (seen in `snap list`)?

kyrofa:
Also, can I see the output of `sudo snap get nextcloud ports`, please?

extraymond:
1. I'm on google compute engine with ubuntu 16.04 64bit.
2. When I try to refresh nextcloud to the above channel, the error indicates that the refresh is not completed, so snap list shows the origin installed channel.

I think since the refresh is not completed, snap get won't get the new option ports.

kyrofa:
Oh interesting, I'm testing on amd64 as well. Would you mind showing me the complete refresh command as well as the error, please?

extraymond:
I'm using the exact command you given above:

|``` sudo snap install nextcloud --channel=stable/pr-349 ``` |



Kyle Fazzari 於 2017年09月13日 00:26 寫道:

kyrofa:
And the error?

kyrofa:
(note that the command you provided is the install command, not refresh)

extraymond:
I tried the refresh one. 
(Edited. Was copying the wrong line from your comment for reply.)

extraymond:
Here's the error log.

```
error: cannot perform the following tasks:
- Run configure hook of "nextcloud" snap if present (run hook "configure": "" is not a valid HTTP port)

```

kyrofa:
Ah! Now that's useful feedback! That's a problem with this PR, fixing now.

extraymond:
Thx! Looking forward to set nextcloud behind caddy!

kyrofa:
Fixed in the PR, new snaps are building now, I'll ping you when they're ready and you should actually be able to refresh and test this.

extraymond:
Thx! I'll give it a try and see if that's working.

Kyle Fazzari 於 2017年09月13日 00:40 寫道:

kyrofa:
@extraymond alright, the amd64 snap has been updated, please try that refresh again.

extraymond:
Hi! I've successful change the port to my like and it's working if I 
connect to the ip:port. But when I set it behind caddy proxy server, it 
redirect to "http://localhost:3000/index.php/apps/files/" which is odd.

I have had success configuring snap service like wekan or rocketchat 
behind proxy.

I think it would need another config like change root-url to the one I 
set in dns server.

Kyle Fazzari 於 2017年09月13日 01:54 寫道:

extraymond:
I got it working! It seems that I forget to change overwritehost via occ.

I think this may be value information for people using reverse proxy.
Exposing host config  via snap interface might be a good option too.

kyrofa:
@extraymond can you describe exactly what you had to do (including commands)? I can put a blurb in the README.

extraymond:
Alright!

So in the nextcloud manaul it states [that](https://docs.nextcloud.com/server/11/admin_manual/configuration_server/config_sample_php_parameters.html#proxy-configurations)

> The automatic hostname detection of Nextcloud can fail in certain reverse proxy and CLI/cron situations. This option allows you to manually override the automatic detection; for example www.example.com, or specify the port www.example.com:8080

So in order to make reverse proxy work (whether nginx or caddy), we would have to overwrite the host to hour desired one by -

```
nextcloud.occ config:set overwritehost --value="example.host"
```

kyrofa:
Excellent, @extraymond you've been so helpful! Does [this blurb](https://github.com/nextcloud/nextcloud-snap/pull/349/files#diff-04c6e90faac2675aa89e2176d2eec7d8R72) sound about right?

extraymond:
It's good I think. You are awesome too.

Kyle Fazzari 於 2017年09月13日 23:43 寫道:

kyrofa:
Tested scenarios:
  - Install from `stable/pr-349`. 
    - Verify ports are set to 80/443 by default by running `sudo snap get nextcloud ports`
    - Change HTTP port with `sudo snap set nextcloud ports.http=81`. Verify that it's no longer available on port 80. Visit port 81, verify that it's working.
    - Verify setting back to port 80 also works.
    - Enable HTTPS with `sudo nextcloud.enable-https self-signed`.
      - Verify that HTTP (port 80) redirects to HTTPS (port 443).
      - Change HTTP port to port 81.
      - Verify that port 81 still redirects to HTTPS (port 443).
      - Change HTTPS port to 444
      - Verify that HTTPS is no longer available on 443. Visit over 444, verify it's working.
      - Verify that port 81 still redirects to HTTPS, but now on port 444. Required a new private browser window, which cached previous redirect to port 443.
      - Verify that ports can be set back to default.

  - Install from stable
    - Refresh to `stable/pr-349`, verify that install succeeds
    - Verify that ports can be changed. No need to be as verbose as the previous test, but test both HTTP and HTTPS.

Everything looks good from the manual tests, and this PR also introduces most of those tests in CI and they're also green. Landing this.

a-m-s:
I installed snap 11.0.5snap1, and followed the "set" instruction, but the open port has not changed. The configure hook appears to run "successfully". The "snap get" command returns the value I set. I tried disable and enable, but still no joy. Am I doing something wrong?

This is running snap on Arch, but that shouldn't matter, right? Nextcloud is running just fine, albeit on port 80.

kyrofa:
Theoretically, no, OS shouldn't matter as long as snapd runs. I'll have to get an arch VM setup and poke at this. Would you mind logging an issue, please?

a-m-s:
Done: #366

a-m-s:
BTW, in case you're not familiar with Arch, I suggest installing Antergos instead. Installing plain Arch is not exactly zero effort.

kyrofa:
Thanks for the heads up! Thankfully, I work on Linux for a living. Arch is already installed and I'm testing :wink: .

PR #375, opened 2017-10-18
php: support changing memory limit

### LINKED ISSUE #374: php: allow memory limit configuration
Now that we have a suitable solution for configuration, the next item on the list is to make the gallery work a little better by allowing PHP's memory limit to be bumped up in a maintainable fashion. It's unfortunate that the gallery requires this for larger pictures, but there's really no way around it.

kyrofa:
Currently PHP uses the default memory limit of 128M. However, the gallery can require more than this for generating previews of large images.

This PR resolves #374 by adding support for changing the memory limit via the `configure` hook-- the `php.memory-limit` option, specifically. It also adds a set of tests verifying this capability.


kyrofa:
Please test this PR by using the `stable/pr-375` branch:

    $ sudo snap install nextcloud --channel=stable/pr-375

    $ sudo snap refresh nextcloud --channel=stable/pr-375

kyrofa:
Been running this in production for several days now. Seems to work well.

Dmitra:
I'm testing Nextcloud capabilities.
Have a directory with 1M+ files.
Tried to index files with files:scan but got error.
First time I tried to increase the memory limit to 1024M it was successfull, I got files indexed. Mysql db size is now 176Mb.
But to open that dir through web interface 1G RAM is not enough. Second increase to 3000M (of 3.5G avail) succeeded too but than after all night swapping due to memory lack the limit became 128M and cannot be changed any more.

Now I removed files with ssh rm -rf, but  nextcloud.occ files:scan fails due to memory allocation.

While this command "sudo snap get nextcloud php.memory-limit" yields 2048M I keep getting error "Allowed memory size of 134,217,728 bytes exhausted".

Is there a way to check where this limit should be applied and why it is not?
Thanks.

PR #397, opened 2017-12-14
php: enable pcntl

### LINKED ISSUE #396: PHP: Enable PCNTL
"The process control (PCNTL) extensions are required in case you want to interrupt long running commands - see http://php.net/manual/en/book.pcntl.php" warnings are annoying, and oftentimes lead people into thinking something is wrong, when it's not.

kyrofa:
This PR resolves #396, getting rid of warnings when using `occ`, which were annoying and often misled people into thinking something was wrong.

Please test this PR using the `stable/pr-397` channel:

    $ sudo snap install nextcloud --channel=stable/pr-397

    $ sudo snap refresh nextcloud --channel=stable/pr-397

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -173,6 +173,7 @@ parts:
       - --with-mcrypt
       - --enable-exif
       - --enable-intl
+      - --enable-pcntl
       - --with-jpeg-dir=/usr/lib
       - --disable-rpath
     stage-packages:


kyrofa:
@pachulo when you're able, mind taking this for a spin?

pachulo:
Hey! Did the refresh and it went OK. I then executed the command to update the database to utf8mb4:

`sudo nextcloud.occ config:system:set mysql.utf8mb4 --type boolean --value="true"`

And no warnings were showed. Do you need any more tests?

kyrofa:
Awesome, thanks @pachulo! No, as long as things still seem to behave okay, that's all I needed.

PR #416, opened 2018-01-14
Configuration of Listen address implemented through hook

### LINKED ISSUE #331: How to get httpd listening to 127.0.0.1 instead of 0.0.0.0?
Hi, for some reason, I have to set redirection from 0.0.0.0:443 to 127.0.0.1:443, which means nextcloud-snap could not use 0.0.0.0:443 but 127.0.0.1:443. How should I do it? I saw only port settings mentioned here.

asmaack:
Implementes/Fixes #331 
https://github.com/nextcloud/nextcloud-snap/issues/331

'Configuration of Listen address implemented through configuration hook, like "sudo snap set nextcloud listen-address=127.0.0.1".

Just added a commit with changes to the README.md, hope it can be of use :-)

### DIFF
--- a/src/apache/scripts/httpd-wrapper
+++ b/src/apache/scripts/httpd-wrapper
@@ -20,6 +20,13 @@ else
 	echo "No certificates are active: using HTTP only"
 fi
 
+# If no listen address is specified, listen on all IPv4 addresses (0.0.0.0)
+if [ $(apache_listen_address) ]; then
+    export LISTEN_ADDRESS="$(apache_listen_address)"
+else
+    export LISTEN_ADDRESS="0.0.0.0"
+fi
+
 export HTTP_PORT="$(apache_http_port)"
 export HTTPS_PORT="$(apache_https_port)"
 
--- a/src/apache/scripts/httpd-wrapper
+++ b/src/apache/scripts/httpd-wrapper
@@ -20,13 +20,7 @@ else
 	echo "No certificates are active: using HTTP only"
 fi
 
-# If no listen address is specified, listen on all IPv4 addresses (0.0.0.0)
-if [ $(apache_listen_address) ]; then
-    export LISTEN_ADDRESS="$(apache_listen_address)"
-else
-    export LISTEN_ADDRESS="0.0.0.0"
-fi
-
+export LISTEN_ADDRESS="$(apache_listen_address)"
 export HTTP_PORT="$(apache_http_port)"
 export HTTPS_PORT="$(apache_https_port)"
 
--- a/src/apache/utilities/apache-utilities
+++ b/src/apache/utilities/apache-utilities
@@ -52,6 +52,21 @@ apache_pid()
 	fi
 }
 
+apache_listen_address()
+{
+    address="$(snapctl get listen-address)"
+    if [ -z $address ]; then
+        address=""
+    else
+        echo "$address"
+    fi
+}
+
+apache_set_listen_address()
+{
+    snapctl set listen-address="$1"
+}
+
 apache_http_port()
 {
 	port="$(snapctl get ports.http)"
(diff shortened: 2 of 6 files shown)


src/apache/scripts/httpd-wrapper
```diff
@@ -20,6 +20,13 @@ else
 	echo "No certificates are active: using HTTP only"
 fi
 
+# If no listen address is specified, listen on all IPv4 addresses (0.0.0.0)
+if [ $(apache_listen_address) ]; then
+    export LISTEN_ADDRESS="$(apache_listen_address)"
+else
+    export LISTEN_ADDRESS="0.0.0.0"
+fi
```
kyrofa: Please follow the pattern created by the port configuration. Have the default in the Apache utilities instead of putting the logic here.

src/apache/utilities/apache-utilities
```diff
@@ -52,6 +52,21 @@ apache_pid()
 	fi
 }
 
+apache_listen_address()
+{
+    address="$(snapctl get listen-address)"
+    if [ -z $address ]; then
+        address=""
```
kyrofa: I'd like to see something more along [these lines](https://github.com/nextcloud/nextcloud-snap/pull/416/files#diff-0b502a9dcd7c3bf1d2d17b3d3ce358d1R74) in this case.

src/hooks/bin/configure
```diff
@@ -16,6 +16,33 @@
 . $SNAP/utilities/php-utilities
 . $SNAP/utilities/hook-utilities
 
+handle_apache_listen_address_config()
+{
+    listen_address="$(apache_listen_address)"
+    
+    # If no Listen address is submitted, then nothing to be done
+    if [ -z "$listen_address" ]; then
```
kyrofa: Please save the default, and also save old values. Then compare against the old values, like everything else in here.

bithex:
ok, I'll do some changes.
It's also possible I was not testing correctly with IPv6, I was using the auto-config IPv6 on the server for testing. I'll try to have the IPv6 logic included with the changes.

bithex:
Just added 2 more commits. Code is now similar to ports config hook.

kyrofa:
@bithex I'm very sorry for the delay here, I've just been caught up in so many other things, here!

The code looks excellent, but the indentation is off, and it needs tests. Have you checked [those out](https://github.com/nextcloud/nextcloud-snap#hacking) at all?

bithex:
I did some testing in our own environment, snap on ubuntu 16.04 server with multiple IPv4 addresses. 
- tested configuring different single IPv4 addresses for listening
- tested changing ports
- tested implementing https/443
- tested implementing letsencrypt certificates
- tested configuring different single Ipv4 addresses after letsencrypt cert installed
- NB! tested configuring invalid IPv4/IPv6 addresses, the regex will reject them and the configuration is not changed, but the error message is not echoed to the stdout. Perhaps you could see why that is happening. Somehow, I find the snap environment is rather difficult to debug.

Clients were Firefox and Chrome, server with basic Nextcloud setup (no extra apps installed on server), all working well with valid IP4 addresses. 

Testing on IPv6 was done with curl local on the IPv6 server, which is very limited, needs more testing in an end-to-end IPv6 environment, just the same, changing between valid IPv6 listen addresses is successful.

As it is, the snap can be configured to listen on one IP address or all (0.0.0.0 or ::)

bithex:
ok, thanks!

naggie:
Looking forward to this -- I'm running an nginx based reverse proxy for all my home services, and wish to listen on a (HTTP) port bound to localhost only, doing ssl-termination/virtualhosting centrally with nginx.

srottschaefer:
What needs to be done to get this merged. The reverse proxy configuration could profit from this as you don't leave your Nextcloud ports exposed that are only meant for the proxy to connect to.

kyrofa:
I'm afraid I don't have time to resolve conflicts and write proper tests for this. I'll close for now. If someone wants to see this through, feel free to make the required changed and re-propose.

bor8:
Workaround here (including user friendly firewall): https://github.com/nextcloud/nextcloud-snap/issues/134#issuecomment-575904440

Issue #426, opened 2018-01-24
Remove extra semicolon from hsts header


bn4t:
The removal of the extra semicolon is suggested by hstspreload.org

![hstspreload.org](https://user-images.githubusercontent.com/17193640/35219519-4616597a-ff73-11e7-90d7-23b7ac1f2330.PNG)

PR #428, opened 2018-01-24
apache: remove extra semicolon from hsts header

bn4t:
Solves #426


kyrofa:
Thaaat's what I'm talking about, nice work. Thanks for sticking with it!

Once the tests are green it goes in.

bn4t:
Thank you for being so patient :)

PR #449, opened 2018-02-13
Update disable-https

19wolf:
grammar fix


kyrofa:
Thanks for the PR, @19wolf! However, PRs need to be to develop, not to master (see the contribution guide). Beyond that, I don't actually think there's a grammar error here-- both seem completely valid. They can also both seem TO BE completely valid!

PR #461, opened 2018-03-01
fix: Add xvfb to test dependencies in README

r4co0n:
Tests could not be run as described in README.md under Hacking, as
a dependency, xvfb, was missing. Now added.

Closes nextcloud/nextcloud-snap#460

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,11 @@ dependencies:
     $ cd tests/
     $ bundle install
 
+Additionally, if you do not have X configured, install the following for a
+'fake' X server.
+
+    $ sudo apt install xvfb
+
 Make sure the snap has a user called "admin" with password "admin" (used for
 login tests):
 


kyrofa:
Thanks for the fix, @r4co0n. Note that I have a few questions on the issue itself that may require tweaks to this.

README.md
```diff
@@ -146,7 +146,7 @@ suite of acceptance tests against it. The tests are written in ruby, using
 capybara and rspec. To run the tests, you first need to install a few
 dependencies:
 
-    $ sudo apt install gcc g++ make qt5-default libqt5webkit5-dev ruby-dev zlib1g-dev
+    $ sudo apt install gcc g++ make qt5-default libqt5webkit5-dev ruby-dev xvfb zlib1g-dev
```
kyrofa: As we discussed in the issue, mind breaking `xvfb` off and saying below this line something like "If you don't have X running, you'll need a virtual framebuffer as well: (provide the install command)."

r4co0n:
As a side note: There's surely a way to tell CircleCI to ignore certain files, isn't there? It would seem appropriate to not do a rebuild on pull requests only changing documentation. (This might also include other static files not relevant to the snap.)

kyrofa:
Probably, but it doesn't matter-- I can merge while the tests are running if I want :stuck_out_tongue: .

PR #462, opened 2018-03-01
apache: remove 'includeSubDomains' from HSTS header

### LINKED ISSUE #435: Disable HSTS after certifcate install nextcloud.enable-https lets-encrypt

bn4t:
Fixes #435

### DIFF
--- a/src/apache/conf/ssl.conf
+++ b/src/apache/conf/ssl.conf
@@ -161,6 +161,6 @@ SSLRandomSeed connect file:/dev/urandom 512
 
     # Enable HSTS only if requested
     <IfDefine EnableHSTS>
-        Header always set Strict-Transport-Security "max-age=63072000; includeSubdomains"
+        Header always set Strict-Transport-Security "max-age=63072000"
     </IfDefine>
 </VirtualHost>


r4co0n:
The `Fixes …` part in the title is not parsed and picked up by GitHub, you should put it in the description, I suppose.

bn4t:
Fixed it. Thank you for the heads up :)

kyrofa:
It actually should be part of both the description and the commit message (from which the default description is generated anyway). That way we know both in review and after the fact the issue that this was supposed to fix. Mind amending this commit?

bn4t:
done :)

kyrofa:
Hmm.... I'm starting to rethink this one as I research it further. Check out [this material](https://www.owasp.org/index.php/HTTP_Strict_Transport_Security_Cheat_Sheet) from owasp:

> Cookies can be manipulated from sub-domains, so omitting the "includeSubDomains" option permits a broad range of cookie-related attacks that HSTS would otherwise prevent by requiring a valid certificate for a subdomain. Ensuring the "Secure Flag" is set on all cookies will also prevent, some, but not all, of the same attacks. 

Knowing that, I doubt this is something we want to just remove for everyone (I certainly want it). Sounds like we need more fine-grained control over this, which makes it quite a bit more work I'm afraid. I'm sorry @bn4t, this PR was nice and simple! Let's chat more about this on the issue, I'll close this for now.

PR #466, opened 2018-03-03
Drop all copy plugin calls, replace with dump calls

r4co0n:
For the boost part, we only stage the 'boost/' subdirectory of the
source archive, as it contains the only files needed for building
mysql, and I could not a find a way to organize the whole archive,
including its 'boost/' folder, to 'boost/' in the snap, short of using
another part.

What i wanted: * -> boost/ <> boost/boost/version.hpp et al exist
(Everything in the archive should reside under directory boost/)
What happened: * -> boost/ <> boost/version.hpp exists
(Everything in the archive, except for the contents of archive folder
boost/ got moved under directory boost/)
Now: boost -> boost/boost/ <> boost/boost/version.hpp et al exist
(Only the directory boost/ from the archive gets moved to boost/boost,
where it is traditionally expected in the snap.)

It should be noted that the dump calls can be further simplified by
reorganizing the src/ folder of the repository:
* Rename scripts/ subfolder to bin/
* Move custom scripts residing in root of src/<part>/ to bin/
* Move custom support files to support-files/
And drop corresponding organize lines.

Closes nextcloud/nextcloud-snap#59


kyrofa:
> It should be noted that the dump calls can be further simplified by
reorganizing the src/ folder of the repository:
>
> - Rename scripts/ subfolder to bin/
> - Move custom scripts residing in root of src// to bin/
> - Move custom support files to support-files/
> And drop corresponding organize lines.

Yeah if we're going to the effort to replace these, might as well go all the way with it as you rightly called out. Are you okay making those changes here?

r4co0n:
I tried to implement a variant of the symlink approach I described above, but unfortunately the dump plugin is still confused about symlinks. I opened [a bug](https://bugs.launchpad.net/snapcraft/+bug/1754010) about this.
When consulting [some](https://en.wikipedia.org/wiki/Symlink_race) [ancient](http://doc.cat-v.org/plan_9/4th_edition/papers/lexnames) [texts](https://news.ycombinator.com/item?id=3075408), I found hints that symlinks are a Bad Idea™ - The only alternative I can see is to just move the files, which will create problems with some users' scripts.

kyrofa:
> Then users could be advised to migrate without breaking every script they wrote that somehow relies on the location.

I'm not really following you here. This is one of the beautiful things of using snaps-- there should be no such scripts to worry about since this is all self-contained. Do you know of something specific?

r4co0n:
I am thinking about people modifying the snap source or parsing a config file they expect to be somewhere. Like, grep through apache config to get listen ports and cross-check this with actually listening ports and expected configuration. Or anything else along those lines...

kyrofa:
Yeah, don't worry about that. It's:

1. Not something that anyone is doing of which I'm aware
2. Not something we want to support for exactly your hesitation
3. Not something that actually works, because files in the snap are read-only and don't change (e.g. the Apache config doesn't change when ports are updated, that's all done with `snap set` and should be checked with `snap get`).

Besides all that, where the files reside in this repository don't directly correspond to where they reside in the built snap. I see no reason to change the latter, we're just reorganizing the former.

snap/snapcraft.yaml
```diff
@@ -211,44 +208,41 @@ parts:
...
   boost:
-    plugin: copy
+    plugin: dump
     source: https://github.com/kyrofa/boost_tarball/raw/master/boost_1_59_0.tar.gz
     source-checksum: sha1/5123209db194d66d69a9cfa5af8ff473d5941d97
-    files:
-      '*': boost/
+    source-subdir: boost/
+    organize:
+      '*': boost/boost/
```
kyrofa: Blech, looks like you hit a gross bug. Rather than force things to continue working this way, I suggest we remove the `source-subdir` and `organize` here completely, and simply change mysql to look for it straight in the staging area by updating [this line](https://github.com/nextcloud/nextcloud-snap/pull/466/files#diff-5fde7a6d86053f0e1d88c0a2a238941fR249) to use `$SNAPCRAFT_STAGE` instead of `$SNAPCRAFT_STAGE/boost`.

kyrofa: No... scratch that, then files start interfering. Leave this the way it is.

kyrofa:
If you don't mind squashing this up into one nice commit, I'll get it merged.

snap/snapcraft.yaml
r4co0n: Now that you went over this again: The way it is now, snapcraft fails to do a rebuild without a full clean. If I go `snapcraft clean --step stage` afterwards and build again, it fails with some `boost/boost/boost/boost/boost/ recursion limit`.
Do you want me to open a bug on Launchpad for this? Otherwise, if you give me a bug number, I might be able to add my observations.

r4co0n: We should be fine with:
```
source-subdir: boost/
organize:
  '*': boost/
```
Or drop `source-subdir` and ignore all files in the archive except `boost/`.

Both approaches would need the WITH_BOOST build parameter to be updated, as you mentioned.

Another simple way would be to call the target directory for the boost libs anything but `boost/`, e.g. `mysql_boost`, this should allow for saner approaches when organizing.

r4co0n:
@kyrofa: I left some more comments in code review, I don't know if this changes your opinion about the fitness of this pull request.

snap/snapcraft.yaml
kyrofa: What... really? You should log a bug, actually, this is dumb. I think I like your last idea best-- let's work around this "folder is the same name is the same name" thing and just call it "mysql_boost".

kyrofa: But please add a comment in the YAML saying why.

r4co0n:
I used a slightly different approach for the boost part. Only stage the boost/ directory, don't organize anything, exclude everything in prime. Therefore, DWITH_BOOST is now using $SNAPCRAFT_STAGE without any subdirectory. The build of MySQL is only interested in the boost/ directory under DWITH_BOOST anyways. This way we don't stage something we don't need and still circumvent the 
bug of not being able to rerun snapcraft without cleaning all steps.

I forgot to open a bug for what we found and will amend the commit again (I mention it in the comment, but left a placeholder there). **Please wait with merging until I have done so.**

r4co0n:
@kyrofa, this should be ready for merge now.

kyrofa:
@r4co0n looks good, although there's a conflict. Mind taking a look?

r4co0n:
@kyrofa, the conflict should be resolved.

PR #482, opened 2018-03-20
Update README.md

uiharu-s:
To `overwritehost`, we need `nextcloud.occ config:system:set`.
`config:set` is not defined.


kyrofa:
Ah, right you are, thank you! However, this pull request needs to target `develop`, not `master`. Mind changing it?

PR #484, opened 2018-03-20
occ: add apices for arguments

3v1n0:
Without this we end up in this issue:
 - https://github.com/nextcloud/server/issues/4291


kyrofa:
Oh man, this is old. We probably want "$@", don't you agree?

3v1n0:
Agree... And done.

kyrofa:
Sorry, just noticed this was going to master-- please squash into a single commit and target to develop instead.

kyrofa:
I'll go ahead and close this, thanks @3v1n0.

PR #485, opened 2018-03-21
occ: add apices and use  use $@ to pass arguments

3v1n0:
Without this we end up in this issue:
 - https://github.com/nextcloud/server/issues/4291


3v1n0:
Would you mind backporting this to stable too :)

kyrofa:
Oh it will, `master` is the release branch. This will be included in the next release.

PR #544, opened 2018-05-02
Update PHP version to 7.1.17

### LINKED ISSUE #537: Update PHP to 7.1.17
This is a security release, so we should upgrade:
http://www.php.net/ChangeLog-7.php#7.1.17

bn4t:
Didn't tested it locally since this is only a security update. Fixes #537.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -156,8 +156,8 @@ parts:
 
   php:
     plugin: php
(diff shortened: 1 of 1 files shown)


kyrofa:
Not 7.1.27, but 7.1.17. Mind amending your commit? Also, please put the `Fixes` on a line by itself, something more like this:

```
php: update to 7.1.17

Fix #537
```

PR #546, opened 2018-05-02
[WIP] Implement http2

bn4t:


bn4t:
Yes I saw the issue. 
I suppose it isn't as bad as back then, since a lot of big sites (Stackoverflow, Cloudflare, Google,...) now adopted http/2.

bn4t:
Hmm. There are also other voices which say http/2 is indeed an improvement....
https://blog.httpwatch.com/2015/01/16/a-simple-performance-comparison-of-https-spdy-and-http2

I could try to implement it as an option. This way both sides would be pleased.

PR #580, opened 2018-05-16
Make Cronjob interval adjustable

### LINKED ISSUE #350: Enhance customization CHRON.PHP allowing spinning down hard drives
The current Cron-job configuration is set by default 15 min. to reschedule a job. This keeps constantly spinning the hard drives in idle as well.
This cause unnecessary spinning of hard drives in the Nextcloud Box, which wastes energy and reduce time life of hard disks.
Would be great to have the possibility to customize the Cron-job configuration to set different time intervals. Additionally, could be great to have to stablish different in a different time range. E.g. Schedule corn-job 10 min interval from 08:00 - 18:00, and 2 hours interval from 18:00 - 08:00.
Any other suggestion to solve this issue would be great. Thanks.

bn4t:
Fixes #350

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,21 @@ To set it to be unlimited (not recommended), use -1:
     $ sudo snap set nextcloud php.memory-limit=-1
 
 
+#### Cronjob interval configuration
+
+By default the cronjob interval is 15 minutes.
+
+To adjust it (say, 10 minutes) simply run:
+
+    $ sudo snap set nextcloud nextcloud.cron-interval=10m
+
+If you want to disable the cronjob completely, run:
+
+    $ sudo snap set nextcloud nextcloud.cron-interval=-1
+
+To reenable it again simply set the `nextcloud.cron-interval` snap variable to a value that isn't `-1`
+
+
 ### Included CLI utilities
 
 There are a few CLI utilities included:
--- a/src/hooks/bin/configure
+++ b/src/hooks/bin/configure
@@ -11,6 +11,10 @@
 # - php.memory-limit (string)
 #   Memory limit for PHP workers. String can include units, e.g. 32M is 32
 #   megabytes.
+#
+# - nextcloud.cron-interval (string)
+#   Interval of the internal Nextcloud cronjob
+#
 
 # shellcheck source=src/apache/utilities/apache-utilities
 . "$SNAP/utilities/apache-utilities"
@@ -18,6 +22,8 @@
 . "$SNAP/utilities/php-utilities"
 # shellcheck source=src/hooks/utilities/hook-utilities
 . "$SNAP/utilities/hook-utilities"
+# shellcheck source=src/nextcloud/utilities/nextcloud-utilities
+. $SNAP/utilities/nextcloud-utilities
 
 handle_apache_port_config()
 {
@@ -83,9 +89,32 @@ handle_php_memory_limit()
 	php_set_previous_memory_limit "$memory_limit"
 }
 
+handle_cronjob_interval()
+{
+	interval="$(cronjob_interval)"
+	previous_interval="$(previous_cronjob_interval)"
+
+	# If no changes were requested, then there's nothing to do here.
+	if [ "$interval" = "$previous_interval" ]; then
+		return 0
+	fi
+
+	# Validate time interval seconds (s), minutes (m), hours (h), "-1" or no specified type(by default seconds) are allowed
+	if ! expr "$interval" : '^\(\(\([0-9]*\.\)\?[0-9]\+[smhd]\?\)\|\(\-1\)\)$' > /dev/null; then
+		echo "\"$interval\" is not a valid timespan" >&2
+		return 4
+	fi
+
+	set_cronjob_interval "$interval"
+	set_previous_cronjob_interval "$interval"
+
+	# restart cronjob
+	snapctl restart nextcloud.nextcloud-cron
+}
+
 # Signal to services that the configure hook is running. Useful to ensure
 # services don't restart until the configuration transaction has completed.
 set_configure_hook_running
 trap 'set_configure_hook_not_running' EXIT
 
-handle_apache_port_config && handle_php_memory_limit
+handle_apache_port_config && handle_php_memory_limit && handle_cronjob_interval
--- a/src/nextcloud/bin/nextcloud-cron
+++ b/src/nextcloud/bin/nextcloud-cron
@@ -11,6 +11,11 @@ done
 printf "done\n"
 
 while true; do
+	# Exit the entire job if the requested interval is -1
+	if [ "$(cronjob_interval)" = "-1" ]; then
+		exit 0
+	fi
+
 	run-php "$SNAP/htdocs/cron.php"
-	sleep 15m
+	sleep "$(cronjob_interval)"
 done
--- a/src/nextcloud/utilities/nextcloud-utilities
+++ b/src/nextcloud/utilities/nextcloud-utilities
@@ -2,6 +2,7 @@
 
 export NEXTCLOUD_CONFIG_DIR="$SNAP_DATA/nextcloud/config"
 export NEXTCLOUD_DATA_DIR="$SNAP_COMMON/nextcloud/data"
+DEFAULT_CRONJOB_INTERVAL="15m"
 
 nextcloud_is_configured()
 {
@@ -17,4 +18,31 @@ wait_for_nextcloud_to_be_configured()
 		done
 		printf "done\n"
 	fi
+}
+
+cronjob_interval()
+{
+	interval="$(snapctl get nextcloud.cron-interval)"
+	if [ -z "$interval" ]; then
+		interval="$DEFAULT_CRONJOB_INTERVAL"
+		set_cronjob_interval "$interval"
+		set_previous_cronjob_interval "$interval"
+	fi
+
+	echo "$interval"
+}
+
+previous_cronjob_interval()
+{
+	snapctl get private.nextcloud.cron-interval
+}
+
+set_cronjob_interval()
+{
+	snapctl set nextcloud.cron-interval="$1"
+}
+
+set_previous_cronjob_interval()
+{
+	snapctl set private.nextcloud.cron-interval="$1"
 }
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -77,7 +77,7 @@ apps:
   nextcloud-cron:
     command: nextcloud-cron
     daemon: simple
-    restart-condition: always
+    restart-condition: on-failure
     plugs: [network, network-bind, removable-media]
 
   # Command for manually installing instead of visiting site to create admin.


src/hooks/bin/configure
```diff
@@ -80,9 +85,34 @@ handle_php_memory_limit()
...
+	interval="$(cronjob_interval)"
+	previous_interval=$CRONJOB_INTERVAL
+
+	# If no changes were requested, then there's nothing to do here.
+	if [ "$interval" = "$previous_interval" ]; then
+		return 0
+	fi
+
+	# Validate time interval seconds (s), minutes (m), hours (h), "-1" or no specified type(by default seconds) are allowed
+	if ! expr "$interval" : '^-\?[0-9]\+[smh]\?$' > /dev/null; then
```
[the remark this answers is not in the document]
bn4t: Thank you for the heads up. Fixed it now.
I think I'll just leave the `0` be, since it doesn't brekas anything.

r4co0n:
I thought some more about this, is there any mechanism in place to avoid a race condition here? We should only allow one cronjob to run at once, I am not sure how this is implemented in Nextcloud's `cron.php`. I suppose we land in [this branch](https://github.com/nextcloud/server/blob/master/cron.php#L136-L151) of the code, don't we?

The old implementation made "sure" the jobs had finished by letting enough time elapse, but suppose someone configures `0` here, maybe just a mistake that went unnoticed.

kyrofa:
Thanks for this, @bn4t. I need to finish up #563 before I land any changes to shell scripts though, so hang tight. I'll give this a closer look soon.

kyrofa:
@bn4t alright, #563 has landed, which caused some conflicts here. Please merge with master when you're able.

r4co0n:
@bn4t, you need to get the latest `devel` branch. After that, you can rebase your work on it using `git rebase devel my_feature_branch`.

Looking at what `cron.php` does, i found [this](https://github.com/nextcloud/server/blob/master/lib/private/BackgroundJob/JobList.php#L207-L210), which implements a check that skips jobs that are currently being run. So you're right, this is already handled.

r4co0n:
Disregard my last comment, if I'm reading what @kyrofa wrote again, I get it as: "Please base your pull request on branch `master`". But this seems to contradict the [contribution guidelines](https://github.com/nextcloud/nextcloud-snap/blob/master/CONTRIBUTING.md), so please wait for clarification by the maintainer.
Sorry for my lack of reading comprehension when writing the last post, I hope I didn't confuse you too much ;)

bn4t:
@r4co0n No problem. Happens to the best of us. :)

r4co0n:
@bn4t, ha, I think I know what's going on now.

The last time, when you wanted your code in `patch-6` based on latest `develop`, you used a merge, which creates a **separate** commit from what's in `develop`, making the branches diverge. So you need to undo that merge.
I think I would create a 'helper branch' from `patch-6`, then rewind `patch-6` to before [605f18f](https://github.com/bn4t/nextcloud-snap/commit/605f18f0c96abc394582c7157d80d66c494b2293), then rebase (as previously described) against `develop` and merge the commits after `605f18f` from your helper branch.

kyrofa:
Yeah listen to @r4co0n, I must have been mostly asleep when typing my "master" comment. Indeed, I meant "develop".

bn4t:
Ok, I think I got it now. Feel free to tell me if messed something up ;)

README.md
```diff
@@ -101,6 +101,21 @@ To set it to be unlimited (not recommended), use -1:
...
 
+#### Cronjob interval configuration
+
+By default the cronjob interval is 15 minutes.
+
+To adjust it (say, 10 minutes) simply run:
+
+    $ sudo snap set nextcloud cron.interval=10m
+
+If you wanna disable the cronjob completely, run:
```
kyrofa: Please let's use proper grammar here: "want to"

src/hooks/bin/configure
```diff
@@ -83,9 +89,34 @@ handle_php_memory_limit()
 	php_set_previous_memory_limit "$memory_limit"
 }
 
+handle_cronjob_interval()
+{
+	interval="$(cronjob_interval)"
+	previous_interval=$CRONJOB_INTERVAL
```
kyrofa: From what I can see, this will never be defined. Is there a reason you didn't follow a model similar to that given in the other configurable items (saving the previously-set interval, etc.)?

src/nextcloud/bin/nextcloud-cron
```diff
@@ -11,6 +11,17 @@ done
...
 
 while true; do
-	run-php "$SNAP/htdocs/cron.php"
-	sleep 15m
+
+
+  # retrieve the current cronjob interval if the environement variable isn't set
+	if [ -z "$CRONJOB_INTERVAL" ]; then
+		CRONJOB_INTERVAL="$(cronjob_interval)"
+		export CRONJOB_INTERVAL
```
kyrofa: It would be cleaner to always have the cronjob interval set and forgo the use of environment variables completely. Just set it to a default value on first run, like the other configurable items.

kyrofa:
Also, check the PR/commit title, "intervall" has one L.

src/hooks/bin/configure
bn4t: Not really. Thought it was simpler. Going to change it to model the other configurable items.

bn4t:
Ok, added the requested changes. Let me know if something is missing.

bn4t:
Squashed the commits into one. I think this is ready to merge. @kyrofa

src/hooks/bin/configure
```diff
@@ -11,13 +11,19 @@
 # - php.memory-limit (string)
 #   Memory limit for PHP workers. String can include units, e.g. 32M is 32
 #   megabytes.
+#
+# - cron.interval (string)
```
kyrofa: Can we please make this `nextcloud.cron-interval`?

src/hooks/bin/configure
```diff
@@ -83,9 +89,35 @@ handle_php_memory_limit()
...
+	if ! expr "$interval" : '^\(\(\([0-9]*\.\)\?[0-9]\+[smhd]\?\)\|\(\-1\)\)$' > /dev/null; then
+		echo "\"$interval\" is not a valid timespan" >&2
+		return 4
+	fi
+
+	set_cronjob_interval "$interval"
+	set_previous_cronjob_interval "$interval"
+	
+	# restart cronjob
+	snapctl restart nextcloud.nextcloud-cron
```
kyrofa: Good find here, I lost track of the release containing this feature.

src/nextcloud/bin/nextcloud-cron
```diff
@@ -11,6 +11,10 @@ done
...
 
 while true; do
-	run-php "$SNAP/htdocs/cron.php"
-	sleep 15m
+
+	if [ "$(cronjob_interval)" != "-1" ]; then
+	  run-php "$SNAP/htdocs/cron.php"
+	fi
+
+	sleep "$(cronjob_interval)"
```
kyrofa: In the README you state that "-1" is a supported arg. What happens here if "-1" is supplied? Errors and a very fast infinite loop taking CPU? If we want to support disabling this job we should probably just exit 0 and change the `restart-condition` to `on-failure`.

src/nextcloud/utilities/nextcloud-utilities
```diff
@@ -18,3 +19,31 @@ wait_for_nextcloud_to_be_configured()
...
 }
+
+
+
+cronjob_interval()
+{
+	interval="$(snapctl get cron.interval)"
+	if [ -z "$interval" ]; then
+		interval="$DEFAULT_CRONJOB_INTERVAL"
+		set_cronjob_interval $interval
```
kyrofa: We need to set the previous one here as well, or the first time the configure hook is run (even for configuring something else) the cron job will be started over.

src/nextcloud/utilities/nextcloud-utilities
```diff
@@ -1,7 +1,8 @@
 #!/bin/sh
 
-export NEXTCLOUD_CONFIG_DIR="$SNAP_DATA/nextcloud/config"
-export NEXTCLOUD_DATA_DIR="$SNAP_COMMON/nextcloud/data"
+export NEXTCLOUD_CONFIG_DIR=$SNAP_DATA/nextcloud/config
+export NEXTCLOUD_DATA_DIR=$SNAP_COMMON/nextcloud/data
```
kyrofa: Please leave these quoted.

src/nextcloud/utilities/nextcloud-utilities
```diff
@@ -18,3 +19,31 @@ wait_for_nextcloud_to_be_configured()
 		printf "done\n"
 	fi
 }
+
```
kyrofa: Please limit line breaks to one between functions.

src/hooks/bin/configure
```diff
@@ -83,9 +89,35 @@ handle_php_memory_limit()
...
+}
+
+
+
+
```
kyrofa: Please remove all but one of these line breaks.

src/nextcloud/utilities/nextcloud-utilities
bn4t: Good catch!

bn4t:
I think it's quite clean if we just exit the cron loop using `break` if the interval is set to `-1` (implemented in 58a14ca).
Or is there a better way I'm not aware of?

src/nextcloud/bin/nextcloud-cron
```diff
@@ -11,6 +11,13 @@ done
...
 
 while true; do
-	run-php "$SNAP/htdocs/cron.php"
-	sleep 15m
+
+  # exit cron loop if interval is "-1"
+	if [ "$(cronjob_interval)" != "-1" ]; then
+	  run-php "$SNAP/htdocs/cron.php"
+	else
+	  break
```
kyrofa: I think this whole loop could be simplified into:

```bash
while true; do
    # Exit the entire job if the requested interval is -1
    if [ "$(cronjob_interval)" = "-1" ]; then
        exit 0
    fi

    run-php "$SNAP/htdocs/cron.php"
    sleep "$(cronjob_interval)"
done
```

bn4t: Didn't think about that. This is definitely a better solution.

src/nextcloud/bin/nextcloud-cron
```diff
@@ -11,6 +11,11 @@ done
...
 while true; do
-	run-php "$SNAP/htdocs/cron.php"
-	sleep 15m
+    # Exit the entire job if the requested interval is -1
+    if [ "$(cronjob_interval)" = "-1" ]; then
+        exit 0
+    fi
+
+    run-php "$SNAP/htdocs/cron.php"
+    sleep "$(cronjob_interval)"
```
kyrofa: Looks good, but please leave these as tabs.

src/nextcloud/utilities/nextcloud-utilities
```diff
@@ -18,3 +19,31 @@ wait_for_nextcloud_to_be_configured()
...
+cronjob_interval()
+{
+	interval="$(snapctl get nextcloud.cron-interval)"
+	if [ -z "$interval" ]; then
+		interval="$DEFAULT_CRONJOB_INTERVAL"
+		set_cronjob_interval "$interval"
+	fi
+
+	set_previous_cronjob_interval "$interval"
```
kyrofa: This needs to be under the `if`. Please refer to other configurable items for how this is done, for example [apache](https://github.com/nextcloud/nextcloud-snap/blob/master/src/apache/utilities/apache-utilities#L61).

src/nextcloud/bin/nextcloud-cron
bn4t: I must have messed this up when copy pasting it. It's now fixed.

PR #608, opened 2018-06-12
mysql: don't consider mysql available until after restart

### LINKED ISSUE #607: wait_for_mysql should wait until mysql has been upgraded, as well
In order to upgrade the mysql schema, mysql itself needs to be run, then restarted a few moments after. Unfortunately, if things run fast enough, the rest of the snap (which typically waits for mysql to be up and running) will march on, running e.g. a Nextcloud upgrade, only to have mysql restart on it after ITS upgrade process has finished.

MySQL should not be considered ready until its past the point of potentially restarting.

kyrofa:
The process of actually firing MySQL up includes upgrading its schema, after which it's restarted. If other daemons fire up at the right time, they could already be relying on MySQL's presence when it restarts out from under them.

This PR fixes #607 by ensuring that other daemons don't consider MySQL ready until after it has
completed this logic.

### DIFF
--- a/src/mysql/bin/start_mysql
+++ b/src/mysql/bin/start_mysql
@@ -12,6 +12,8 @@ if mysqld --initialize-insecure --basedir="$SNAP" --datadir="$SNAP_DATA/mysql" -
 	new_install=true
 fi
 
+set_mysql_setup_running
+
 # Start mysql
 "$SNAP/support-files/mysql.server" start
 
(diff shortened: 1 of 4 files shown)


kyrofa:
Please test this PR by using the `beta/pr-608` channel. Watch the PHP journal as it's initializing, and compare it to what you see in `stable` today: in `stable` you see MySQL initializing while PHP tries to forge ahead, whereas in this PR you'll see PHP politely wait until MySQL has restarted once.

PR #618, opened 2018-06-19
nextcloud: update php to 7.2.8

### LINKED ISSUE #617: Update PHP to 7.2.x
It seems that Nextcloud 13.x supports PHP 7.2.x:

- https://docs.nextcloud.com/server/13/admin_manual/installation/system_requirements.html

And as per this comparison, it seems that the overall performance should be a little bit better:

- https://www.phoronix.com/scan.php?page=news_item&px=PHP-7.2-Benchmarks-RC6

bn4t:
Fixes #617


pachulo:
Hey! Thanks for this @bn4t !

It did not pass the tests, so maybe you need to do a modification like this one to the `extension_dir`:

- https://github.com/nextcloud/nextcloud-snap/pull/399/files#diff-87ecae46e1b62c47ad6f7d057766a815L724

Using the folder for PHP 7.2: `no-debug-non-zts-20170718`

bn4t:
Thank you for the tip @pachulo. 
Was wondering why it was failing...

pachulo:
You're welcome! Let's see if it did the trick!

bn4t:
It worked! Is there something else to do before this can be merged?

bn4t:
Tested it on ubuntu 16.04. Everything seems to work fine.

pachulo:
Maybe we should add this PHP extension: https://github.com/php/pecl-encryption-mcrypt (as per https://github.com/nextcloud/nextcloud-snap/issues/617#issuecomment-399096692) and see it that application works.

pachulo:
Please, update this to 7.2.7 too @bn4t !

bn4t:
Seems like `phpize` is executed in the wrong directory. 
Not sure what the `source-subdir` option from the php plugin does. Can we set the execution directory for `phpize` with this? @kyrofa

snap/snapcraft.yaml
```diff
@@ -208,6 +206,9 @@ parts:
       # Build the redis PHP module
       - source: https://github.com/phpredis/phpredis/archive/3.1.4.tar.gz
         source-checksum: sha256/656cab2eb93bd30f30701c1280707c60e5736c5420212d5d547ebe0d3f4baf71
+      # Build the mcrypt PHP module  
+      - source: https://pecl.php.net/get/mcrypt-1.0.1.tgz
```
pachulo: I think that with the custom php plugin that @kyrofa did, you need to use the repository, not the pecl module. Something like:

`- source: https://github.com/php/pecl-encryption-mcrypt.git`

Should be enough in this case.

snap/snapcraft.yaml
```diff
@@ -195,7 +194,6 @@ parts:
       - libpng12-dev
       - libjpeg9-dev
       - libbz2-dev
-      - libmcrypt-dev
```
pachulo: I think that we need to keep this.

kyrofa:
@bn4t any chance you tested this with the user_saml like @pachulo suggested?

bn4t:
Haven't had the time so far. Might be able to do it today.

bn4t:
Is this ready to merge or is there anything still missing? @pachulo

snap/snapcraft.yaml
```diff
@@ -208,6 +207,8 @@ parts:
       # Build the redis PHP module
       - source: https://github.com/phpredis/phpredis/archive/3.1.4.tar.gz
         source-checksum: sha256/656cab2eb93bd30f30701c1280707c60e5736c5420212d5d547ebe0d3f4baf71
+      # Build the mcrypt PHP module  
+      - source: https://github.com/php/pecl-encryption-mcrypt.git
```
kyrofa: Can you please change this to a specific release, along with a checksum, please?

tests/spec/enable_http2_spec.rb
```diff
@@ -0,0 +1,11 @@
+feature "Enabling HTTPS" do
```
kyrofa: This probably doesn't belong in this PR.

bn4t: Not sure how that happened. I removed it now. Thanks for pointing it out.

snap/snapcraft.yaml
bn4t: I changed now the source to the latest commit and added checksum verification.

bn4t:
The test currently fails because the httpd download link(http://ftp.wayne.edu/apache/httpd/httpd-2.4.33.tar.bz2) returns a `404` error. 

@kyrofa A PR of yours (#646) fixes this. Should I just wait until #646 is merged and the merge the `development` branch into `update-php-to-7.2.x`? Or is there a better solution?

kyrofa:
@bn4t yeah hold off, I'll land that today.

pachulo:
Please @bn4t update this to PHP 7.2.8
@kyrofa should we open an issue & PR for PHP 7.1.20?
Both 7.2.8 & 7.1.20 are security releases:

- http://www.php.net/ChangeLog-7.php#7.1.20
- http://www.php.net/ChangeLog-7.php#7.2.8

bn4t:
@pachulo Thanks for the heads up. I updated it now to `7.2.8`.

kyrofa:
I've been doing a little research on this one, and I'm not really liking the move to 7.2 right now. Folks are discussing only making apps work with 7.1, etc. until they have time to move away from mcrypt, so I see two possible paths:

1. Apps start checking PHP version and start saying they don't work with 7.2
2. We now start supporting a legacy and seemingly unmaintained PHP extension that deals with security.

Neither of those is good news for snap users if we move to 7.2. I think the smarter play right now is to stick to PHP 7.1 until apps move away from mcrypt. Thoughts?

pachulo:
@kyrofa regarding **2**: How is that different from keep on using PHP 7.1.x? I mean, that version includes mcrypt and I guess that it's based on the extension...is it better because the PHP team would fix any possible problem?
In my understanding is that mcrypt is not bad as per se, but that it's complex to use and it gives developers who use it a million ways to shoot in their foot:

- https://secure.php.net/manual/en/migration71.deprecated.php

Apart from that, I would like to measure the speed gainings with PHP 7.2 to see if the update is worthwhile.

kyrofa:
@pachulo yeah, in 7.1 they're still on the hook for supporting it, whereas in 7.2 I suspect it will languish and die. I'm not sure we'll ever be able to remove it in the snap if we support it now.

pachulo:
OK @kyrofa then I'm with you that the most sensible choice is to keep the snap out of the PHP 7.2 loop till enough apps migrate to use sodium instead of mcrypt and then update.

kyrofa:
Alright, I'll close this for the time-being, then, but thank you for doing this @bn4t-- we'll revive it when the time is right.

JamesBream:
+1 for 7.1 from me. The performance increases that 7.2 brings, while nice, are fairly modest in controlled tests so probably won't translate to much of a real-world improvement for us (though I'd be really interested to see the figures if you do any testing @pachulo).

According to [this](http://php.net/supported-versions.php), 7.1 has security support until December 2019 so I don't think there's any urgency to update on that front.

The PECL extension hasn't been touched since 2016 and libMcrypt hasn't been updated since 2007 - not ideal for a security library.

Ubuntu LTS has just moved to 7.2 as the default PHP version with 18.04, as have other distros recently, which I think will begin to force app devs to prioritise dropping Mcrypt so I don't think we'll have to hold out on 7.1 for long. Perhaps we could revisit this with NC14 - which itself might also encourage apps to be updated?

kyrofa:
Exactly my thinking, @JamesBream. I suspect this won't be an issue for very long, and we can roll out 7.2 once mcrypt isn't needed anymore. I really think it's just user_saml, so let's keep an eye on it.

pachulo:
Let's hope so @kyrofa @JamesBream ! Thanks a lot for this @bn4t !

PR #619, opened 2018-06-20
apache: add TLS settings

### LINKED ISSUE #616: SSL Encryption Settings
Hi there,

I've like to disable TLS 1.0 and only use TLS 1.1 and TLS 1.2.
All i could find was a Wiki entry from 2017 saying I should edit /var/snap/nextcloud/current/apache/config/ but that folder and its content seem to have vanished since the Wiki entry was last edited.

What can I do to edit these settings?

bn4t:
Fixes #616 

I think the simplest (and also most user friendly) way to implement TLS settings is to specify two "modes" (old and modern). 
In this PR this is done using the `apache.oldtls` setting. 

Feel free to suggest changes or improvements to this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,21 @@ the automatic detection (including the port if necessary), e.g.:
     $ sudo nextcloud.occ config:system:set overwritehost --value="example.com:81"
 
 
+#### TLS configuration
+
+By default apache accepts the Tlsv1, Tlsv1.1 and Tlsv1.2 versions and a cipher list that is compatible with most browsers.
+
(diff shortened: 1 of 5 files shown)


bn4t:
Tested it on ubuntu 16.04. Everything seems to be working fine.

kyrofa:
Thanks for doing this, @bn4t. However, this is not something I'd like to make configurable right now. If we did end up doing that, it would be more cipher suite specific, but these days it may be a good idea to disable TLS 1.0 and perhaps 1.1 across the board. I'd like to see what ssltest says about both of those (disabling 1.0 and disabling both 1.0 and 1.1), feel like testing it out?

PR #629, opened 2018-07-07
Add Imagemagick support in PHP 

### LINKED ISSUE #592: imagemagick not installed? (favicon)
Hi, I'm running nextcloud snap 13.0.2 stable snap, and saw that i cant have the favicon auto-generated, because imagemagick is not installed (or enabled?)
Is there a way to fix that in snap?

errorfrom the theming plugin: Install the Imagemagick PHP extension with support for SVG images to automatically generate favicons based on the uploaded logo and color.

pachulo:
With the imagick PHP extension and it's dependencies.

When merged, this should fix #592

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -183,14 +183,12 @@ parts:
       # These are only included here until the OS snap stabilizes
       - libxml2
       - libpng12-0
-      - libmagickwand-6.q16-2
-      - libmagickcore-6.q16-2
+        #- libmagickwand-6.q16-2
+        #- libmagickcore-6.q16-2
     build-packages:
       - libxml2-dev
       - libcurl4-openssl-dev
       - libpng12-dev
-      # This causes a conflict with libmagick* deps, that need libjpeg8-dev
-      #- libjpeg9-dev
       - libbz2-dev
       - libmcrypt-dev
       - libldap2-dev
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -180,11 +180,11 @@ parts:
       - --with-libdir=lib/ARCH_TRIPLET
       - --with-ldap
     stage-packages:
+      - libmagickwand-6.q16-2
+      - libmagickcore-6.q16-2
       # These are only included here until the OS snap stabilizes
       - libxml2
       - libpng12-0
-        #- libmagickwand-6.q16-2
-        #- libmagickcore-6.q16-2
     build-packages:
       - libxml2-dev
       - libcurl4-openssl-dev
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -189,7 +189,8 @@ parts:
       # These are only included here until the OS snap stabilizes
       - libxml2
       - libpng12-0
-      - imagemagick
+      - libmagickwand-6.q16-2
+      - libmagickcore-6.q16-2
     build-packages:
       - libxml2-dev
       - libcurl4-openssl-dev
@@ -199,7 +200,8 @@ parts:
       - libmcrypt-dev
       - libldap2-dev
       - libfreetype6-dev
-      - libmagickwand-dev
+      - libmagickwand-6.q16-dev
+      - libmagickcore-6.q16-dev
     prime:
      - -sbin/
      - -etc/
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -195,15 +195,14 @@ parts:
       - libxml2-dev
       - libcurl4-openssl-dev
       - libpng12-dev
-      - libjpeg9-dev
+      # This causes a conflict with libmagick* deps, that need libjpeg8-dev
+      #- libjpeg9-dev
       - libbz2-dev
       - libmcrypt-dev
       - libldap2-dev
       - libfreetype6-dev
       - libmagickwand-6.q16-dev
       - libmagickcore-6.q16-dev
-      # Don't really know why this dependency cannot be met
-      - libjpeg8-dev
     prime:
      - -sbin/
      - -etc/
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -210,7 +210,7 @@ parts:
       - source: https://github.com/phpredis/phpredis/archive/3.1.4.tar.gz
         source-checksum: sha256/656cab2eb93bd30f30701c1280707c60e5736c5420212d5d547ebe0d3f4baf71
       # Build the imagick module
-      - source: https://github.com/mkoppanen/imagick/
+      - source: https://github.com/mkoppanen/imagick.git
         source-tag: 3.4.3
 
   redis:
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -189,6 +189,7 @@ parts:
       # These are only included here until the OS snap stabilizes
       - libxml2
       - libpng12-0
+      - imagemagick
     build-packages:
       - libxml2-dev
       - libcurl4-openssl-dev
@@ -208,6 +209,9 @@ parts:
       # Build the redis PHP module
       - source: https://github.com/phpredis/phpredis/archive/3.1.4.tar.gz
         source-checksum: sha256/656cab2eb93bd30f30701c1280707c60e5736c5420212d5d547ebe0d3f4baf71
+      # Build the imagick module
+      - source: https://github.com/mkoppanen/imagick/
+        source-tag: 3.4.3
 
   redis:
     plugin: redis
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -199,6 +199,7 @@ parts:
       - libmcrypt-dev
       - libldap2-dev
       - libfreetype6-dev
+      - libmagickwand-dev
     prime:
      - -sbin/
      - -etc/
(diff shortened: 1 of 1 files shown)


kyrofa:
There was quite a security issue with Imagemagick there for a while, have you looked into that? Are there still concerns?

pachulo:
Well, it has a pretty bad record regarding security vulns:
- https://www.cvedetails.com/vulnerability-list/vendor_id-1749/Imagemagick.html
- http://changelogs.ubuntu.com/changelogs/pool/main/i/imagemagick/imagemagick_6.8.9.9-7ubuntu5.11/changelog

And though I have not spent time trying to find out the real risk of having this activated I rely on the opinion of experts:
- https://lcamtuf.blogspot.com/2016/05/clearing-up-some-misconceptions-around.html

>  And so, the advice sorely missing from the "ImageTragick" webpage is this:
>  * If all you need to do is simple transcoding or thumbnailing of potentially untrusted images, don't use ImageMagick. Make a direct use of libpng, libjpeg-turbo, and giflib; for a robust way to use these libraries, have a look at the source code of Chromium or Firefox. The resulting implementation will be considerably faster, too.
>   * If you have to use ImageMagick on untrusted inputs, consider sandboxing the code with seccomp-bpf or an equivalent mechanism that robustly restricts access to all user space artifacts and to the kernel attack surface. Rudimentary sandboxing technologies, such as chroot() or UID separation, are probably not enough.
>   * If all other options fail, be zealous about limiting the set of image formats you actually pass down to IM. The bare minimum is to thoroughly examine the headers of the received files. It is also helpful to explicitly specify the input format when calling the utility, as to preempt auto-detection code. For command-line invocations, this can be done like so:
>     * convert [...other params...] -- jpg:input-file.jpg jpg:output-file.jpg
>     * The JPEG, PNG, and GIF handling code in ImageMagick is considerably more robust than the code that supports PCX, TGA, SVG, PSD, and the likes.

But I'm not really sure if this is a stopper for the integration.

kyrofa:
Good news is, seccomp-bpf is already used by virtue of it being a snap. I'm not sure that will stop the denial of service issues, though. This makes me nervous, grr, I'm having second thoughts now. @pachulo how do you feel about this? @r4co0n how about you?

ykgmfq:
Over in Docker camp: https://github.com/nextcloud/docker/issues/263

pachulo:
Well, I think that taking into account that:
- In the default snap install this should only be used for generating some stuff when using a custom theme.
- We will have the packets upgraded when new fixes are released, as they are in the main archive of Ubuntu.

I would say that we can take the risk of having these inside the snap.

brainfrz82:
If imagemagick is included, how would this impact preview generation? On one hand, imagemagick should be a bit faster on preview generation, but on the other hand imagemagick supports a lot of new filetypes. I'm running on a pentium, but it would take weeks to generate previews for the 1TB+ RAW files i'm storing :) Couldn't even make an estimation on how a rPi would handle this.

snap/snapcraft.yaml
```diff
@@ -189,15 +189,20 @@ parts:
       # These are only included here until the OS snap stabilizes
       - libxml2
       - libpng12-0
+      - libmagickwand-6.q16-2
+      - libmagickcore-6.q16-2
```
kyrofa: Have you tried leaving these off and seeing if the snapcraft CLI pulls them over for you?

snap/snapcraft.yaml
```diff
@@ -189,15 +189,20 @@ parts:
...
     build-packages:
       - libxml2-dev
       - libcurl4-openssl-dev
       - libpng12-dev
-      - libjpeg9-dev
+      # This causes a conflict with libmagick* deps, that need libjpeg8-dev
+      #- libjpeg9-dev
```
kyrofa: You can safely remove this comment, although I appreciate seeing it once to describe why the change was necessary, it doesn't need to stay in the YAML.

snap/snapcraft.yaml
pachulo: No, I haven't, but I will now.

snap/snapcraft.yaml
```diff
@@ -183,15 +183,18 @@ parts:
       # These are only included here until the OS snap stabilizes
       - libxml2
       - libpng12-0
+        #- libmagickwand-6.q16-2
+        #- libmagickcore-6.q16-2
```
kyrofa: Were you able to verify that the snap can still use magick with these gone?

pachulo: No, but I'm compiling snaps with this code right now and let you know!

steiny2k:
Hi what's the status with this change? I'm especially curious whether it would also support the newly added .heic files which we added to nextcloud/core with https://github.com/nextcloud/server/issues/7406 I guess the snap would need libheif as well to make it work. Would you be interested if I dig into this?

kyrofa:
Last I knew @pachulo was still [playing with the stage-packages](https://github.com/nextcloud/nextcloud-snap/pull/629#pullrequestreview-141331009). @pachulo is this good to go?

pachulo:
Well, the last thing I've tried is what I wrote here https://github.com/nextcloud/nextcloud-snap/issues/592#issuecomment-410550426 : everything seemed right, but I was not able to really see it working.

JamesBream:
This change does make me uneasy particularly since this PR looks to be using imagemagick's default minimum security policies (correct me if I'm wrong @pachulo ) which Imagemagick themselves advise are not suitable for anything pubic facing.

Snap sandboxing has been brought up which I suppose does provide the system with some protection from being totally compromised, but wouldn't protect anything within the snap (e.g. sensitive user data) from being compromised (again, correct me if I'm wrong).

Others here might disagree with me but I question the benefit of such a change given Imagemagick's history, simply for autogenerating favicons and preview generation of non web-safe image formats? Also, if we do properly secure imagemagick, that typically means restricting image formats it will process which will likely mean that the only real benefit will be to auto-generate favicons?

I'm open to being persuaded on this one, but it seems like we're adding a pretty big attack surface here for all users of the snap so definitely worth the discussion.

pachulo:
Hey @JamesBream ! First, thanks for the feedback!!

I don't know what those "imagemagick's default minimum security policies" are, so I cannot answer that.
Regarding the other questions:
1. Yes, an exploitation of imagemagick would allow an attacker to access sensitive data stored in nextcloud.
2. Yes, maybe the functional gain is too low compared to the risks.

So @kyrofa do we ditch this then?

JamesBream:
@pachulo Oops, I meant to include a link to https://www.imagemagick.org/script/security-policy.php

pachulo:
OK, now I get it. I only have to say that all your points are quite fair, so I have no arguments to try to convince you for adding this to the snap.

brainfrz82:
I'm with steiny2k on this one. I also run my nextcloud for my family. I don't think enterprise users would/should generally run the snap version. Maybe the imagemagick module could be implemented as an opt-in setting via OCC?

r4co0n:
The concerns raised here seem to apply upstream as well. I can't see [netcloud/vm](https://github.com/nextcloud/vm) mitigating any of this, though they are [installing php-imagick](https://github.com/nextcloud/vm/blob/1592fb45e504158a393970fd7231f9d070647df6/nextcloud_install_production.sh#L194).

Digging through the Debian packages, I couldn't find anything about `policymap`, so there are apparently no restrictions applied as outlined on imagick.org . But maybe I didn't look hard enough.

kyrofa:
I'm sorry for the delay everyone, I've been thinking a lot about this. @brainfrz82 my wife is a professional photographer, so believe me when I say that I want RAW previews as well. However, I just don't feel right pushing a piece of software that we know has a shoddy security record to all 17k snap users. I know of businesses running this as well, but even if all snap users were home users, I don't want to open them up to that kind of risk either.

With that said, if someone can think of a way to make this risk opt-in, let me know (I can think of none, I'm afraid, since such a decision would seem to be needed at build-time). It would be nice to see Nextcloud move to using something with a better record, instead.

@pachulo I truly appreciate the effort you've put into this. I hope you understand.

brainfrz82:
@kyrofa In hindsight, with nextcloud's subtitle being "Protecting your data", I do believe you made the right call here. Practicality should not trump security if security is the main thing you advertise with :)

Let's hope they'll think of something more secure upstream :)

pachulo:
No problem @kyrofa !

r4co0n:
I had the pleasure to talk about this with Nextcloud's groundskeeper (he said ['Hausmeister'](https://en.wiktionary.org/wiki/Hausmeister), groundskeeper Willy from the Simpsons is 'Hausmeister Willy' in German) @MorrisJobke about this, as he happened to talk at our company stuff about handling git properly.

I was told that the security-critical features of ImageMagick are mostly about complex image formats, think SVG and its embedded JavaScript. Rendering of those formats is disabled in Nextcloud by default, which should make these concerns a non-issue. Someone should go have a look at the upstream code to find the portion where they handle those file formats and make sure everything indeed is as I claim it to be, I might do this myself, but wanted to report my finding anyway.

This also concerns nextcloud/vm#743

MorrisJobke:
There is also an alternative approach: https://github.com/nextcloud/server/issues/13551

MorrisJobke:
See also https://github.com/nextcloud/server/pull/12821#issuecomment-451120426 and the comment from @kyrofa - maybe we should re-evaluate this a bit more.

pachulo:
Hey, it seems like upstream is solving this, at least for the `image/*` preview generation, in Nextcloud 24: https://github.com/nextcloud/server/pull/24166

:partying_face:

kyrofa:
Good find, @pachulo. And given that it's an external service, it looks like nothing extra will be needed in the snap at all for it to work, you just need the service.

PR #714, opened 2018-09-15
nextcloud: update to 14.0.3

r4co0n:
Update Nextcloud to version 14.0.0 . This follows the latest
upstream release.


r4co0n:
This works in a local build, updating from the latest stable release.

kyrofa:
The tests are not happy with this. I'll take a closer look.

r4co0n:
I thought of releasing this in the same vein as you. I bet people would have sooner or later asked about the new version. This makes them aware that we got the update on our radar and gives them the chance to contribute by testing a beta release once we get the automatic tests fixed.

Tests fail with 
```
Capybara::Webkit::JavaScriptError:
{:line_number=>7, :message=>"SyntaxError: Unexpected token '>'",
:source=>"http://localhost/apps/notifications/js/notifications.js?v=821eafa6-0"}
```
Strangely, my local `/apps/notifications/js/notifications.js` has only 6 lines, with everything but the first being a license comment.

snap/snapcraft.yaml
```diff
@@ -1,5 +1,5 @@
 name: nextcloud
-version: 13.0.5snap1
+version: 14.0.0snap1
```
kyrofa: Mind reverting this? The version gets bumped as part of the release process.

kyrofa:
@r4co0n I pushed some extra fixes for v14, haven't messed with the tests just yet.

kyrofa:
Huh, tests passed locally. Let's see what happens now.

kyrofa:
[This bug](https://github.com/nextcloud/server/issues/11107) is hitting us. I think I'd like to hold off on releasing v14 until it's fixed, or we'll just get more bug reports. Thoughts?

r4co0n:
@kyrofa wrote:
> [This bug](https://github.com/nextcloud/server/issues/11107) is hitting us. I think I'd like to hold off on releasing v14 until it's fixed, or we'll just get more bug reports. Thoughts?

I am with you on this. Do you have a rough overview of other bugs biting v14? Upstream is using the tag [`14-beta-feedback`](https://github.com/nextcloud/server/labels/14-beta-feedback), but it is not used everywhere, at least the bug you found doesn't have the tag.

Right now, there are 22 issues in that list, I haven't seen through them, but they hopefully aren't all release-critical for us. 

If anyone reading this has a v14 Nextcloud running somewhere already, please tell us if you encountered issues and provide relevant bug numbers. (Please open the bugs in [`nextcloud/server`](https://github.com/nextcloud/server/issues) after you've thoroughly verified it is not yet reported.) It is also appreciated if you tell us everything went smooth for you.

pachulo:
I'm also with you @kyrofa . It seems to be always like that, every new major release (X.0.0) never comes without pain.
But I also think that we should be that conservative for the `stable` channel and that 14.0.0 should already be in the `14/candidate`, or at least `14/beta`, channel because that way we would enable adventurous people to test it.

kyrofa:
> [...] 14.0.0 should already be in the `14/candidate` [...]

Great minds, I've pushed 14.0.0snap1-rc1 (this PR) to 14/candidate.

pachulo:
14.0.1 is already out:

- https://github.com/nextcloud/server/releases

The bug is not yet solved https://github.com/nextcloud/server/issues/11107 so, should we update this PR @kyrofa ?

pachulo:
And 14.0.2 is out now. @r4co0n could you update this PR with it?

pachulo:
@r4co0n ping
@kyrofa can you update this? I really think we should have 14.0.2 in the `14/candidate` channel.

kyrofa:
Dang. Github went belly up on me yesterday, things still seem wonky this morning.

kyrofa:
Alright, in candidate now.

mahfoudfx:
I Think you should consider #778 before switching to the v14 stable channel. I had the same issue.

brainfrz82:
At this point in time, wouldn't it be better to wait for NC15 and go straight to that release?

pachulo:
> At this point in time, wouldn't it be better to wait for NC15 and go straight to that release? When there are no blocking issues of course.

I would say no, because the upgrade from one major version from another is never painless, so I can't really imagine skipping a major version and everything working as it should

Besides, nextcloud X.0.0 versions always need some polishing, so I think that we should follow our own update path for the snap.

kyrofa:
Closed in favor of #799.

PR #727, opened 2018-09-20
Fix for issue #708

### LINKED ISSUE #708: php_readline missing;cannot start indexing for fulltextsearch
Hi 

I'm using the latest snap on a Ubuntu 18.4.1 LTS server, trying to setup fulltextsearch-* with elastic search and tesseract for OCR. When I try to start indexing I get the same error as described here:
https://github.com/nextcloud/fulltextsearch/issues/376

The proposed solution is to install php-readline. How can I do that within the snap? I'm guessing I can't, but is there any way to get this up and running?

Thank you

IvyUK:
Included php --with-readline and the associated build package fixes #708

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -179,6 +179,9 @@ parts:
       # Enable ldap.
       - --with-libdir=lib/ARCH_TRIPLET
       - --with-ldap
+      
+      # Enable readline for elastic full text search
+      - --with-readline
     stage-packages:
       # These are only included here until the OS snap stabilizes
       - libxml2
@@ -192,6 +195,7 @@ parts:
       - libmcrypt-dev
       - libldap2-dev
       - libfreetype6-dev
+      - libreadline-dev
     prime:
      - -sbin/
      - -etc/


snap/snapcraft.yaml
```diff
@@ -192,6 +195,7 @@ parts:
       - libmcrypt-dev
       - libldap2-dev
       - libfreetype6-dev
+      - libreadline-dev
```
kyrofa: We're not typically willing to introduce new dependencies into the snap. Finding documentation for how fulltextsearch works is difficult, but I'd be surprised if this was all that was required to enable all its providers. Have you actually built this snap and tested it with various providers?

IvyUK: Yes, I have built the snap and upgraded my install with it. This has resolved the issue. I have the full text search app and the files & elastic search plug in.

https://apps.nextcloud.com/apps/fulltextsearch
https://apps.nextcloud.com/apps/files_fulltextsearch
https://apps.nextcloud.com/apps/fulltextsearch_elasticsearch

IvyUK:
Any thoughts on how we move this forward ?

kyrofa:
> [...] in this case it seems the fix doesn't require all that much.

Every proposal in and of itself falls into that category. It's when taken together that they're not sustainable. I'm sorry folks :frowning_face: . I'll see if I can come up with some text for the README to better communicate this.

PR #729, opened 2018-09-23
apache, php: add debug mode

### LINKED ISSUE #612: Reduce ServerToken Info Output and Disable ServerSignature
The apache server should by default not expose so much detailed information that gives a potential attacker valuable information about the exact versions in use:

> Date: Thu, 14 Jun 2018 15:37:19 GMT
> Server: Apache/2.4.33 (Unix) OpenSSL/1.0.2g
> X-Powered-By: PHP/7.1.17

```
ServerTokens Prod
ServerSignature Off 
```
Should be part of the default configuration especially since they cannot be changed by the user in a snap setup.

kyrofa:
By default, both Apache and PHP provide detailed version information in the HTTP headers and error pages. This PR resolves #612 by making the default behavior hide this information, but allows them to optionally be enabled by adding a new configuration value named `mode` which can contain two possible values: "debug" and "production".

Please test this PR from the `beta/pr-729` channel:

    $ sudo snap install nextcloud --channel=beta/pr-729

Or, if you already have it installed:

    $ sudo snap refresh nextcloud --channel=beta/pr-729


kyrofa:
@pachulo and @r4co0n, I'd like a review from at least one of you on this one. How to test it is your call, but I'll tell you how I tested it (beyond the tests in this PR, of course): `curl -I <nextcloud address>`, which should print the headers. In production mode (which should be the default, but you can also run `snap set nextcloud mode=production`), the "server" header should only be "Apache". There should be nothing at all in the headers about PHP. Also, if you navigate to a random page in Nextcloud such that a 404 is displayed, there should be nothing mentioned about Apache.

In debug mode (`snap set nextcloud mode=debug`), the "server" header should mention both the version of Apache as well as the operating system. There should also be a "x-powered-by" header that mentions the version of Apache. Finally, a 404 page should mention Apache as well.

pachulo:
I've tested this with Firefox dev tools and these are the headers I see for `debug` mode when going to a valid page:

```
Server: Apache/2.4.34 (Unix) OpenSSL/1.0.2g
X-Powered-By: PHP/7.1.22
```

And those when going to an invalid (404) one:

`Server: Apache/2.4.34 (Unix) OpenSSL/1.0.2g`

And for `production` mode either valid and invalid (404) pages show only this:

`Server: Apache`

kyrofa:
Thanks @pachulo, that works for me!

Issue #745, opened 2018-10-08
failed to set datadir to /var/snap/nextcloud/8971/mysql/


samtygier:
After rebooting for a kernel update my nextcloud service will not come back up. I am unable to connect via web browser. Everything had been working fine for months. The main error seems to be in the mysql component

```
Oct 07 22:52:44 lithium nextcloud.mysql[944]: 2018-10-07T21:52:43.804787Z 0 [Warning] TIMESTAMP with implicit DEFAULT value is deprecated. Please use --explicit_defaults_for_timestamp server option (see documentation for more details).
Oct 07 22:52:44 lithium nextcloud.mysql[944]: 2018-10-07T21:52:44.166170Z 0 [ERROR] --initialize specified but the data directory has files in it. Aborting.
Oct 07 22:52:44 lithium nextcloud.mysql[944]: 2018-10-07T21:52:44.166285Z 0 [ERROR] Aborting
Oct 07 22:52:45 lithium nextcloud.mysql[944]: Starting MySQL
Oct 07 22:52:46 lithium mysqld_safe[2053]: Starting mysqld daemon with databases from /var/snap/nextcloud/8971/mysql
Oct 07 22:52:47 lithium mysqld[2058]: /snap/nextcloud/8971/bin/mysqld (mysqld 5.7.22-log) starting as process 2058 ...
Oct 07 22:52:47 lithium audit[2058]: SECCOMP auid=4294967295 uid=0 gid=0 ses=4294967295 subj==snap.nextcloud.mysql (enforce) pid=2058 comm="mysqld" exe="/snap/nextcloud/8971/bin/mysqld" sig=0 arch=c000003e syscall=116 compat=0 ip=0x7f425a712f09 code=0x50000
Oct 07 22:52:47 lithium kernel: audit: type=1326 audit(1538949167.156:27): auid=4294967295 uid=0 gid=0 ses=4294967295 subj==snap.nextcloud.mysql (enforce) pid=2058 comm="mysqld" exe="/snap/nextcloud/8971/bin/mysqld" sig=0 arch=c000003e syscall=116 compat=0 ip=0x7f425a712f09 code=0x50000
Oct 07 22:52:47 lithium mysqld[2058]: failed to set datadir to /var/snap/nextcloud/8971/mysql/
Oct 07 22:52:47 lithium mysqld[2058]: /snap/nextcloud/8971/bin/mysqld: Shutdown complete
[...]

PR #747, opened 2018-10-10
Use full path to mysqld

samtygier:
Fix for #745

### DIFF
--- a/src/mysql/bin/start_mysql
+++ b/src/mysql/bin/start_mysql
@@ -8,7 +8,7 @@ new_install=false
 
 # Make sure the database is initialized (this is safe to run if already
 # initialized)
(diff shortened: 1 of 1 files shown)


samtygier:
On Debian the snap is not confined, so calling mysqld can start /usr/sbin/mysqld causing nextcloud to fail to start #745 

Not tested, but maybe this is along the right lines.

samtygier:
This is not sufficient to fix the issue. The main mysqld process is started from src/mysql/support-files/mysql.server but that looks like it is doing the paths correctly.

kyrofa:
Thanks @samtygier, but the whole point of snaps is to be cross-distro. If they aren't, that's a bug in snapd or snapcraft, not this snap. I'd rather try to figure out what's broken in those then try to work around it here, it doesn't scale.

samtygier:
Thanks, that makes sense. Have you opened a bug in snapd/snapcraft?

PR #766, opened 2018-10-25
Update version to 13.0.7snap1

kyrofa:
Also update README and CHANGELOG.


kyrofa:
Gonna try changing this to use back-merges from master so we can start using `git describe`.

PR #804, opened 2018-12-03
Doc Created: How to Migrate Away from nextcloud-snap

stone212:
`nextcloud-snap` is very good.  But there are use cases when it is not the best solution.  Because of the way the snap is listed on the NC website as the first option many people will use it without knowing this.  Or a user might grow past the snap and want more direct control over the installation.

Because of this I documented my successful migration away from `nextcloud-snap` to Nextcloud server on Ubuntu 16.04 and I request it to be into the `nextcloud-snap` repository to answer these tickets (one is mine):

https://github.com/nextcloud/nextcloud-snap/issues/735
https://github.com/nextcloud/nextcloud-snap/issues/788

### DIFF
--- a/doc/HowTo-migrateFromSnapToManual.md
+++ b/doc/HowTo-migrateFromSnapToManual.md
@@ -12,6 +12,8 @@ The `nextcloud-snap` is very good.  For some situations it is the perfect soluti
 If you need to move from `nextcloud-snap` to Nextcloud server here are some notes:
 
 
+This document makes some assumptions.  It assumes you use `root` user to make changes.  You should not do this if you don't know what you're doing.  Then you will have to add 'sudo' to most of these commands.
+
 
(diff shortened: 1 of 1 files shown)


pachulo:
Thanks a lot for this @stone212 !
Just one thing, don't you think that this belongs to the Wiki?

kyrofa:
I agree that this is wiki material. @stone212 it's open for editing.

stone212:
@kyrofa @pachulo I understand  I added this to the wiki.  Thank you.  And just to say it here: I think you do great work on nextcloud-snap.  I just needed more control over my installation.

kyrofa:
@stone212 thank you for the kind words! We totally get it, the snap has a pretty specific target audience, and it's perfectly reasonable to need more flexibility than it offers.

PR #837, opened 2019-01-10
Release 14.0.5snap1

kyrofa:
- php: update to 7.1.26
- nextcloud: update to 14.0.5
- ci: update rubygems for tests
- ci: enable daily v14 and v15

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,24 +64,6 @@ jobs:
           # Run the tests
           command: ./.circleci/runtests.sh
 
-  test-daily-v12:
-    working_directory: ~/nextcloud-snap
-    machine: true
-    steps:
-      - checkout
-
-      - run:
-          # Install the snap and create an admin user
-          command: |
-            sudo apt update -qq
-            sudo apt install -y snapd
-            sudo snap install nextcloud --channel=12/edge
-            sudo nextcloud.manual-install admin admin
-
-      - run:
-          # Run the tests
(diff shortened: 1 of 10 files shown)


kyrofa:
One last test @pachulo, then we can ship.

kyrofa:
Looks good to me, fresh and upgrade from stable.

pachulo:
Well, I was already in the `14/candidate` channel, so the update happened automatically during night and everything is up and running!

By the way, did you notice that the new version is almost 10 MB trimmer now?
```
channels:                                              
  stable:       13.0.8snap1              (10317) 184MB -
  candidate:    14.0.5snap1-rc1          (10753) 175MB -
```

kyrofa:
Alright this should be rolling to everyone soon.

nehrka:
Just looked at the login screen of my nextcloud and found the maintenance message. When looking into the console I found: 
`$ sudo nextcloud.occ maintenance:mode --off 
The current PHP memory limit is below the recommended value of 512MB.
Nextcloud or one of the apps require upgrade - only a limited number of commands are available
You may use your browser or the occ upgrade command to do the upgrade
Maintenance mode disabled`

After disabling the maintenance mode I found the following message instead of a login screen:
`The files of the app (updatenotification) were not replaced correctly. Make sure it is a version compatible with the server.`

eldertfrancke:
@nehrka for a solution see #843

nehrka:
@eldertfrancke Thanks works like a charm.

PR #840, opened 2019-01-11
Update PHP to the latest 7.2 release

### LINKED ISSUE #617: Update PHP to 7.2.x

pachulo:
Now that we are on Nextcloud 14 we can start to test this.

This PR resolves #617

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -172,8 +172,8 @@ parts:
 
   php:
     plugin: php
-    source: https://php.net/get/php-7.1.26.tar.bz2/from/this/mirror
-    source-checksum: sha256/5b351ca86bc7e4600778aaf1d61ab9e4e38864efa86ab4cc4d5b02ea7f542ae6
+    source: https://php.net/get/php-7.2.14.tar.bz2/from/this/mirror
+    source-checksum: sha256/f56132d248c7bf1e0efc8a680a4b598d6ff73fc6b9c84b5d7b539ad8db7a6597
     source-type: tar
     install-via: prefix
     configflags:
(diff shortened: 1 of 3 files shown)


kyrofa:
This doesn't run, I'm afraid. I suspect you need to update the php.ini, [here](https://github.com/nextcloud/nextcloud-snap/blob/master/src/php/config/php.ini#L726) specifically.

pachulo:
Damn, I always forget about the change in the API date...Thanks!

pachulo:
OK, now the failure is for the test that checks the PHP string in debug mode:
```
Failures:

  1) Change operating mode debug
     Failure/Error: expect(response["x-powered-by"]).to match /PHP\/7\.1\.\d+/
     
       expected "PHP/7.2.14" to match /PHP\/7\.1\.\d+/
       Diff:
       @@ -1,2 +1,2 @@
       -/PHP\/7\.1\.\d+/
       +"PHP/7.2.14"
```
Nothing important I guess...

kyrofa:
Go ahead and fix that test, then-- check for 7.2 instead of 7.1

kyrofa:
Please test this PR using the `beta/pr-840` channel:

    $ sudo snap install nextcloud --channel=beta/pr-840

Or, if you already have it installed:

    $ sudo snap refresh nextcloud --channel=beta/pr-840

kyrofa:
Upstream says user_saml is fixed, and we're not getting any testers that have the app, which means this will need to turn into a bit of a scream test (although I'm pretty confident this is fine these days, I can install the app for example).

PR #973, opened 2019-04-25
nextcloud: update to 16.0.0

### LINKED ISSUE #972: Update nextcloud to v16

kyrofa:
This PR resolves #972 by updating Nextcloud to the latest v16 release.

---

Note that this PR gets the groundwork of upgrading out of the way and gives us something to test, but we'll continue our tradition of not upgrading everyone until the first point release (16.0.1). Please use the `16/candidate` channel if you want to upgrade to 16.0.0 now.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -158,8 +158,8 @@ parts:
 
   nextcloud:
     plugin: dump
-    source: https://download.nextcloud.com/server/releases/nextcloud-15.0.7.tar.bz2
-    source-checksum: sha256/3e6158951fa72010ccd50dbeac05d8df162183f7bbc62a1c6c89ed7081fa9d49
+    source: https://download.nextcloud.com/server/releases/nextcloud-16.0.0.tar.bz2
+    source-checksum: sha256/4532f7028b1d9bf060f75ac4fbbde52a59ecd9c9155f3178a038d3cf3609402e
     organize:
       '*': htdocs/
       '.htaccess': htdocs/.htaccess


kyrofa:
Only issue I've discovered so far (which I discovered in RC2) is not a blocking bug: https://github.com/nextcloud/server/issues/15236 .

kyrofa:
Hmm... something is too old on the Trusty runners used by CircleCI for these tests to pass (they pass locally on Bionic). Trying to figure that out.

omyno:
> Hmm... something is too old on the Trusty runners used by CircleCI for these tests to pass (they pass locally on Bionic).
Is there a reason not to specify bionic in `.circleci/config.yml`?

kyrofa:
> Is there a reason not to specify bionic in `.circleci/config.yml`?

I'd love to, but [the docs](https://circleci.com/docs/2.0/configuration-reference/#machine) lead me to believe such a thing is impossible with the `machine` executor. I've [asked the question](https://discuss.circleci.com/t/machine-executor-still-using-ubuntu-14-04-trusty-tahr/30018), though.

omyno:
Thanks for the reply. It's irritating that they still rely on 14.04 as the default. Apparently you can use also use `image: ubuntu-1604:201903-01`, which in turn uses
> Ubuntu 16.04, docker 18.09.3, docker-compose 1.23.1

Might be worth a try.

kyrofa:
Wait, how did I miss that?! Thanks @omyno, I'll give that a shot now.

kyrofa:
Ah, it seems these tests don't pass in Xenial either. I'll continue trying to move to chrome headless.

kyrofa:
Alright, got things sorted by moving the tests to chrome.

kyrofa:
I have extracted the test changes from this PR into #977 so it can land before this, which will unblock the failing daily tests against upstream Nextcloud's master and allow us to setup daily tests against v16 as well.

PR #1011, opened 2019-05-19
Release 15.0.8snap1

kyrofa:
- certbot: patch out untested chown calls
- nextcloud: update to 15.0.8
- ci: enable daily v16


kyrofa:
Tested fresh install and upgrade from stable. Also re-confirmed that Let's Encrypt certificates can be both obtained and renewed.

R8s6:
@kyrofa what should I do if my nextcloud server is behind an reverse proxy (HAProxy)?

Checking the log I noticed this:

> To fix these errors, please make sure that your domain name was entered correctly and the DNS A/AAAA record(s) for that domain contain(s) the right IP address. Additionally, please check that your computer has a publicly routable IP address and that no firewalls are preventing the server from communicating with the client. If you're using the webroot plugin, you should also verify that you are serving files from the webroot path you provided.


The server is behind a pfsense router, and this pfsense router gets the Let's Encrypt certificate using ACME with Cloudflare DNS. It actually successfully renewed the certificate a month ago. 

The nextcloud runs at the default http/https port behind a reverse proxy (HAProxy).

The nextlcoud server got its 1st certificate this way with no problem 3 months ago and has been working very well.

Under `/var/snap/nextcloud/current/certs/certbot/config/renewal/[mydomain].conf`, i have:

`authenticator = nextcloud:webroot`

Is it because of this? Thanks!

kyrofa:
@R8s6 please log an issue so we can properly discuss.

PR #1054, opened 2019-07-04
Snap usage example for php configuration

eespitiarea:
Given the comments and questions posted in this [issue](https://github.com/nextcloud/nextcloud-snap/issues/77), I think including an example of how to use snap to setup php values would ease configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ snap permission to access removable media by connecting that interface:
 
 Beyond the typical Nextcloud configuration (either by using `nextcloud.occ` or
 editing `/var/snap/nextcloud/current/nextcloud/config/config.php`), the snap
-exposes extra configuration options via the `snap set` command.
+exposes extra configuration options via `snap set` and `snap get` commands.
 
 
 #### HTTP/HTTPS port configuration
@@ -97,6 +97,25 @@ To set it to be unlimited (not recommended), use -1:
 
     $ sudo snap set nextcloud php.memory-limit=-1
 
+#### Uploading big files > 512MB
+
+The default maximum file size for uploads is 512MB. You can increase this limit up to what your filesystem and operating system allows. There are certain hard limits that cannot be exceeded:
+
+- < 2GB on 32Bit OS-architecture
+- < 2GB with IE6 - IE8
+- < 4GB with IE9 - IE11
+
+64-bit filesystems have much higher limits; consult the documentation for your filesystem. By default, the file size limit is set to 512MB, if you'll like to customize this limit, you can run:
+
+    $ sudo snap set nextcloud php.upload-max-filesize=16G
+    $ sudo snap set nextcloud php.post-max-size=16G
+
+Adjust these values for your needs. If you see PHP timeouts in your logfiles, increase the timeout values, which are in seconds:
+
+    $ sudo snap set nextcloud php.max-input-time=3600
+    $ sudo snap set nextcloud php.max-execution-time=3600
+
+The [mod_reqtimeout](https://httpd.apache.org/docs/current/mod/mod_reqtimeout.html) Apache module could also stop large uploads from completing. If you’re using this module and getting failed uploads of large files either disable it in your Apache config or raise the configured RequestReadTimeout timeouts.
 
 #### Cronjob interval configuration
 


kyrofa:
Thank you for working to improve our documentation, however, the documented functionality doesn't actually exist, nor is the unrelated documentation correct (e.g. you can upload > 2GB on 32-bit, try it).

eespitiarea:
@kyrofa, couple of comments:

1. Took the info from https://docs.nextcloud.com/server/14/admin_manual/configuration_files/big_file_upload_configuration.html, does that means that section is outdated or is just not applicable here?
2. How is it that `sudo snap set nextcloud php.upload-max-filesize=16G` does change the settings persistently while using `nextcloud.occ` doesn't?

Appreciate your comments, just trying to understand.

worikgh:
This is effecting me too
I have a snap installation of next cloud
I can upload a 1M file but not a 2M file/.
I cannot find any logs anywhere lohhing errors but using the webdav interface I get `HTTP Error: Request Entity Too Large` error.  
This fits with what apears to be in the php.ini files:
```
# find  /snap/nextcloud/ -name "php.ini" -exec grep -Hni upload_max_filesize '{}' \;
/snap/nextcloud/13968/config/php/php.ini:802:upload_max_filesize = 2M
/snap/nextcloud/14509/config/php/php.ini:802:upload_max_filesize = 2M
```
I ran the command `snap set nextcloud php.upload-max-filesize=16G` to no effect.
Is there really no way to change settings like this?
2M is really very small these days.

kyrofa:
> [...] does that means that section is outdated or is just not applicable here?

Indeed, that's not applicable to the snap.

> How is it that sudo snap set nextcloud php.upload-max-filesize=16G does change the settings persistently while using nextcloud.occ doesn't?

`sudo snap set nextcloud php.upload-max-filesize=16G` does absolutely nothing. It's part of the functionality that, as I mentioned, doesn't actually exist. I have no idea where you got it, or what gives you the impression it has any effect. All the configuration capable via `snap set` is [explicitly documented](https://github.com/nextcloud/nextcloud-snap#configuration).

kyrofa:
@worikgh I've never run into that limit. Any chance you have a proxy or some other piece of tech sitting in front of the snap on the network?

worikgh:
This is effecting me too
I have a snap installation of next cloud
I can upload a 1M file but not a 2M file/.
I cannot find any logs anywhere lohhing errors but using the webdav interface I get `HTTP Error: Request Entity Too Large` error.  
This fits with what apears to be in the php.ini files:
```
# find  /snap/nextcloud/ -name "php.ini" -exec grep -Hni upload_max_filesize '{}' \;
/snap/nextcloud/13968/config/php/php.ini:802:upload_max_filesize = 2M
/snap/nextcloud/14509/config/php/php.ini:802:upload_max_filesize = 2M
```
I ran the command `snap set nextcloud php.upload-max-filesize=16G` to no effect.
Is there really no way to change settings like this?
2M is really very small these days.

worikgh:
I am proxying through nginx.  I see no reason why that should have a effect, but maybe.
Are you saying that you use snap-nextcloud's file uploading interface for large files (over 2M)?

Today I am rebuilding from something other than snap, where I have control over all the parameters. If I remember to I will report I solve the problem.

worikgh:
I have installed next cloud from source in a container
I adjusted the php `upload_max_filesize` to 2000M
Still can not upload a 2M file, but can upload a 1M file.
I am proxying through `nginx` with...
```
        location / {
                 proxy_pass http://<local IP of container>;
                 proxy_set_header Host $host;
        }
```

I do not think the proxying is the problem.  But I have been wrong in the past....

worikgh:
1M seems to be the limit.
I have tried 1.1M (and 2M, 1.5 and 1.25) all have same error

worikgh:
Egg on my face!
The problem *was* the nginx proxy.
Changing the configuration `client_max_body_size` to a large value fixed it.  By default it barfs at 1M
Golly!

kyrofa:
Yeah that's the exact config value I was going to suggest taking a look at. Glad you got it solved.

Issue #1057, opened 2019-07-09
Update to Nextcloud 16.0.3


stondino00:
https://github.com/nextcloud/server/releases/tag/v16.0.3

kyrofa:
Dupe of #972, we don't need a new one for every new v16 release until we're actually ON v16.

PR #1002, opened 2019-05-16
nextcloud: update to 16.0.5

kyrofa:
This PR resolves #972 by updating Nextcloud to the latest v16 release.

PR #1058, opened 2019-07-09
Update Nextcloud 16.0.3

stondino00:
Fixes issue #1057


kyrofa:
This is covered by #1002, let's keep the upgrade to v16 there for continuity.

PR #1070, opened 2019-07-23
mysql: update to 5.7.27

kyrofa:
This PR resolves #1068 by updating MySQL to the latest 5.7 release.

PR #1071, opened 2019-07-23
For 16/candidate users upgrade PHP and MySQL

stondino00:
Individuals on 16/candidate nextcloud 16.0.3, PHP 7.2.20 and MySQL 5.7.27


kyrofa:
Sorry @stondino00 I think we got our wires crossed on #1070. When I said "proposing such a change would have no destination" I meant "such a change cannot land and should not be proposed." I'm spinning up builds for this now, hang tight.

stondino00:
Ok cool.  No worries.  My bad.

PR #1098, opened 2019-08-16
Export/import keys

### LINKED ISSUE #1097: Handle keys and certs in export/import script
The export/import scripts will currently grab encryption keys, but only if a full data export/import is performed.  Let users export/import their keys separately if desired.  While we're at it, we can also handle certs.

Pull request incoming

the-sane:
Fixes #1097

**Some notes:**
I took most of this code directly from my own backup script, which I have tested and does successfully back up keys and certs.  I haven't tested these specific scripts because I have no idea how they get turned into their corresponding binaries in the snap. :woman_shrugging: I mostly just changed variable names for these scripts so it should be fine, but please someone double check my work.

It is backwards compatible with previous backups.  When importing everything from an older backup, if keys aren't found, it just skips them.

When exporting everything, if no keys are found on the system, it skips them so the rest of the backup can proceed normally.

I assume you're going to want to change the location of the temporary files I use in export_keys() and import_keys(), `/tmp/nextcloud_occ.err` and `/tmp/nextcloud_rsync.err`.  I didn't know where snap prefers to put this type of thing.

**Some questions:**
1. Is there a cleaner way to get the status of encryption and list users other than using nextcloud.occ as I do?
2. Does the built-in "run_command" need to be used instead where I execute cat or mkdir?
3. Is "run-command" going to break the piping I use in import-data line 107?

**Testing:**
To test, it should be sufficient to create text files in the appropriate directories:
`/var/snap/nextcloud/current/certs/live/test.cert`
`/var/snap/nextcloud/common/nextcloud/data/files_encryption/OC_DEFAULT_MODULE/test.key`
`/var/snap/nextcloud/common/nextcloud/data/[USER]/files_encryption/OC_DEFAULT_MODULE/test.key`

### DIFF
--- a/src/import-export/bin/export-data
+++ b/src/import-export/bin/export-data
@@ -26,7 +26,7 @@ print_usage()
 	echo "    Export Nextcloud data suitable for migrating servers."
 	echo "    By default this includes the"
         echo "    apps, database, config, keys, and data"
-	echo "    (equivalent to running $COMMAND -abcdef)."
+	echo "    (equivalent to running $COMMAND -abcde)."
 	echo ""
 	echo "Available options:"
 	echo "    -h: Display this help message"
--- a/src/import-export/bin/export-data
+++ b/src/import-export/bin/export-data
@@ -25,7 +25,7 @@ print_usage()
 	echo "    $COMMAND [OPTIONS]"
 	echo "    Export Nextcloud data suitable for migrating servers."
 	echo "    By default this includes the"
-        echo "    apps, database, config, certs, keys, and data"
+        echo "    apps, database, config, keys, and data"
 	echo "    (equivalent to running $COMMAND -abcdef)."
 	echo ""
 	echo "Available options:"
@@ -35,7 +35,6 @@ print_usage()
 	echo "    -c: Include the config"
 	echo "    -d: Include the data (can be quite large)"
 	echo "    -e: Include the encryption keys"
-	echo "    -f: Include the certificates"
 }
 
 export_apps()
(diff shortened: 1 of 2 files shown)


src/import-export/bin/export-data
```diff
@@ -80,20 +83,78 @@ export_data()
...
+		cat "$occ_err"
+	fi
+	if [ "$encryption_status" = "false" ]; then
+		# because these encryption directories do not exist until encryption is enabled,
+		# rsync will throw a bunch of directory not found errors
+		echo "Encryption does not appear to be enabled, skipping."
+	else
+		# back up system-wide master keys, public sharing keys, recovery keys
+		echo "Exporting system-wide keys..."
+		if ! rsync -ah --info=progress2 "$NEXTCLOUD_DATA_DIR"/files_encryption "${backup}"/keys/ 2>>"$rsync_err"; then
```
kyrofa: I'm a smidge uncomfortable with this. Does this directory only ever contain keys?

src/import-export/bin/export-data
```diff
@@ -23,16 +23,19 @@ print_usage()
...
+	echo "    (equivalent to running $COMMAND -abcdef)."
 	echo ""
 	echo "Available options:"
 	echo "    -h: Display this help message"
 	echo "    -a: Include the (non-default) apps"
 	echo "    -b: Include the database"
 	echo "    -c: Include the config"
 	echo "    -d: Include the data (can be quite large)"
+	echo "    -e: Include the encryption keys"
+	echo "    -f: Include the certificates"
```
kyrofa: Can we please split the export of keys and certs into two PRs (and two issues)?

src/import-export/bin/export-data
the-sane: As far as I can tell, yes.  From the [official documentation](https://docs.nextcloud.com/server/16/admin_manual/configuration_files/encryption_configuration.html):

> The encryption keys are stored in the following directories:
> 
> `data/<user>/files_encryption`
>     Users’ private keys and all other keys necessary to decrypt the users’ files
> `data/files_encryption`
>     private keys and all other keys necessary to decrypt the files stored on a system wide external storage 

And even more detail is provided [here](https://github.com/syseleven/nextcloud-tools/blob/master/server-side-encryption.md#file-types) that is waiting to be merged into the official docs.

src/import-export/bin/export-data
the-sane: Oh yes, that would make more sense wouldn't it.  I'm not very advanced with github but I'll give it a go!  This PR will probably end up with some extra commits to delete the certificates bit while I put it in a different PR.  I have no idea how to clean that up.

the-sane: Done.  Let me know if I did that right.

PR #1136, opened 2019-10-03
Release 16.0.5snap2

kyrofa:
- php: update to 7.2.23
- php: check file timestamps for opcache
- tests: disable v14 dailies and enable v16


kyrofa:
Tested fresh install as well as upgrade from stable.

PR #1144, opened 2019-10-17
feat: add imagemagick

mkg20001:
Add imagemagick to fix the warning that the php module is missing

Do I need to add the php module or just the imagemagick package?


mkg20001:
> After much discussion and considering, it has been decided that ImageMagick is just too risky to include for everyone at this time. If there comes a time where we can make it opt-in, we'll do that.

From #592 

Possibly could be enabled/disabled via snapctl, script would then automatically delete/add the config for the php module.

Issue #1161, opened 2019-10-31
Allow max execution time and max input time change via snap


lpalgarvio:
#### Describe the bug
Allow max execution time and max input time to be changed via snap.

#### To Reproduce
Currently using,
1. snap set nextcloud php.max-execution-time="300"
2. snap set nextcloud php.max-input-time="600"
wont achieve any result.

#### Expected behavior
Would expect php-fpm www pool and php.ini configuration to change at least (must research apache as well).

#### OS/snapd/snap version
```
root@nextcloud:~# snap list nextcloud
Name       Version      Rev    Tracking  Publisher   Notes
nextcloud  16.0.5snap3  16739  16        nextcloud✓  -
root@nextcloud:~# snap version
snap    2.42
snapd   2.42
series  16
ubuntu  18.04
kernel  4.4.0-165-generic
```

#### Partial solution
/snap/nextcloud/current/config/php/php-fpm.d/www.conf:
```
-php_value[max_execution_time] = 3600
+php_value[max_execution_time] = ${PHP_MAX_EXECUTION_TIME}
```
```
-php_value[max_input_time] = 3600
+php_value[max_input_time] = ${PHP_MAX_INPUT_TIME}
```

/snap/nextcloud/current/config/php/php.ini:
```
-max_execution_time = 3600
+;max_execution_time = 30
+max_execution_time = ${PHP_MAX_EXECUTION_TIME}
```
```
-max_input_time = 3600
+;max_input_time = 60
+max_input_time = ${PHP_MAX_INPUT_TIME}
```

MJBG:
Hi,
Doesn't work for me. Any news ?

Issue #1160, opened 2019-10-30
Allow upload size limit change via snap


lpalgarvio:
#### Describe the bug
Allow upload limits to be changed via snap.

#### To Reproduce
Currently using,
1. snap set nextcloud php.post-max-size
2. snap set nextcloud php.upload-max-filesize
wont achieve any result.

#### Expected behavior
Would expect php-fpm www pool and php.ini configuration to change at least (must research apache as well).

#### OS/snapd/snap version
root@nextcloud:~# snap list nextcloud
Name       Version      Rev    Tracking  Publisher   Notes
nextcloud  16.0.5snap3  16739  16        nextcloud✓  -
root@nextcloud:~# snap version
snap    2.42
snapd   2.42
series  16
ubuntu  18.04
kernel  4.4.0-165-generic

#### Partial solution
/snap/nextcloud/current/config/php/php-fpm.d/www.conf:
```
-php_value[post_max_size] = 16G
+;php_value[post_max_size] = 16G
+php_value[post_max_size] = ${PHP_POST_MAX_SIZE}
```
```
-php_value[upload_max_filesize] = 16G
+;php_value[upload_max_filesize] = 2M
+php_value[upload_max_filesize] = ${PHP_UPLOAD_MAX_FILESIZE}
```

/snap/nextcloud/current/config/php/php.ini:
```
-post_max_size = 8M
+;post_max_size = 8M
+post_max_size = ${PHP_POST_MAX_SIZE}
``````
-upload_max_filesize = 2M
+;upload_max_filesize = 2M
+upload_max_filesize = ${PHP_UPLOAD_MAX_FILESIZE}
```

lpalgarvio:
#77 
#586

Fuseteam:
bot php.upload-max-filesize and php.post-max-size appear to be configurable via snap set nextcloud

Fuseteam:
> bot php.upload-max-filesize and php.post-max-size appear to be configurable via snap set nextcloud

i stand corrected it was an illusion

ddesmond:
have issues with slower connections and timeouts, does the snap execution time works?
when i try to get the current values it just throws an error like so:
error: snap "nextcloud" has no "php.max-execution-time" configuration option
any hints ?

Fuseteam:
> have issues with slower connections and timeouts, does the snap execution time works?
> when i try to get the current values it just throws an error like so:
> error: snap "nextcloud" has no "php.max-execution-time" configuration option
[...]

PR #1162, opened 2019-10-31
Merge Allow upload size limit change via snap #1160 and Allow max execution time and max input time change via snap #1161

lpalgarvio:
Merges
- Merge Allow upload size limit change via snap #1160
- Allow max execution time and max input time change via snap #1161

Not tested

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,19 @@ To set them to be unlimited (not recommended), use -1:
     $ sudo snap set nextcloud php.max-input-time=-1
 
 
+#### PHP upload limit configuration
+
+By default, PHP will use 8M as the post max size and 2M as the upload max filesize.
+If you notice files are not being uploaded or the uploads being interrupted in your
+Nextcloud log, you may need to set these to a higher value.
+
+If you'd like to set the post max size and upload max filesize to a higher value
+(say, 16G, the default in snap), run:
+
+    $ sudo snap set nextcloud php.post-max-size=16G
+    $ sudo snap set nextcloud php.upload-max-filesize=16G
+
+
 #### Cronjob interval configuration
 
 By default the cronjob interval is 15 minutes.
--- a/README.md
+++ b/README.md
@@ -83,19 +83,35 @@ the automatic detection (including the port if necessary), e.g.:
     $ sudo nextcloud.occ config:system:set overwritehost --value="example.com:81"
 
 
-#### PHP Memory limit configuration
+#### PHP memory and execution configuration
 
 By default, PHP will use 128M as the memory limit. If you notice images not
 getting previews generated, or errors about memory exhaustion in your Nextcloud
 log, you may need to set this to a higher value.
 
-If you'd like to set the memory limit to a higher value (say, 512M), run:
+If you'd like to set the memory limit to a higher value (say, 512M, the default
+in snap), run:
 
     $ sudo snap set nextcloud php.memory-limit=512M
 
 To set it to be unlimited (not recommended), use -1:
 
     $ sudo snap set nextcloud php.memory-limit=-1
+
+By default, PHP will use 30s as the max execution time and 60s as the max input
+time. If you notice images not getting previews generated, or errors about scripts
+being interrupted in your Nextcloud log, you may need to set these to a higher value.
+
+If you'd like to set the max execution time and max input time to a higher value
+(say, 3600s, the default in snap), run:
+
+    $ sudo snap set nextcloud php.max-execution-time=3600
+    $ sudo snap set nextcloud php.max-input-time=3600
+
+To set them to be unlimited (not recommended), use -1:
+
+    $ sudo snap set nextcloud php.max-execution-time=-1
+    $ sudo snap set nextcloud php.max-input-time=-1
 
 
 #### Cronjob interval configuration
--- a/src/hooks/bin/configure
+++ b/src/hooks/bin/configure
@@ -9,8 +9,14 @@
 #   is enabled).
 #
 # - php.memory-limit (string)
-#   Memory limit for PHP workers. String can include units, e.g. 32M is 32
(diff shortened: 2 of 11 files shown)


lpalgarvio:
try build again?

kyrofa:
I don't think one should ever need to play with these settings. We have our own apache; our own php; our own Nextcloud; we should be able to make these settings whatever we want them to be. I can't think of a single reason why anyone wouldn't want to be able to upload large files, so I think we should just automatically unlock the largest file upload we can rather than adding a bunch of knobs.

kyrofa:
I'm going to go ahead and close this, but we can have further discussion about the merits here on the bugs.

lpalgarvio:
Hello @kyrofa,

I'll start with the easy question.

> You wrote all the tests but you haven't run them on your own?

Nope. I'm still getting familiar with snap so I don't know yet how to build snap packages and test them. However the scripts are very straight forward and I've inspected carefully to not make mistakes. I also considered that the CI pipeline will actually run these tests so that also helps a bit. When I get the chance, I will see how to build the thing and actually test it, as I had limited time the other day.

PS: This is also incomplete, because I only implemented the PHP side. Apache configuration also needs adaptation, but I did not want to invest time there till these PHP changes could be approved. Hence if they get in, I'll invest further time on this to develop apache config and experiment with building the snap package on my computer. 

> However, I don't think one should ever need to play with these settings. We have our own apache; our own php; our own Nextcloud; we should be able to make these settings whatever we want them to be. I can't think of a single reason why anyone wouldn't want to be able to upload large files, so I think we should just automatically unlock the largest file upload we can rather than adding a bunch of knobs.

The reasoning for changing upload limits is that it will make sense for many organizations and individuals, and since it's not actually easy or straightforward to change these settings on PHP (with FPM), it's also actually a pain with snap because the binary/config filesystem is readonly, I think it would be of general interest that these come included as customizable config with the package as to promote snap adoption - and not deter it.

It is also non-standard and a bit unpolitical to change things on a readonly filesystem as a workaround for a missing feature upstream that will just frustrate users and sysadmins and make them use different approaches to snap, but this can be avoided by allowing this feature in.

Now to why organizations and individuals might want to change the limit or not want a high limit (unlimited btw as you may know is not possible):

- for certain organizations/people, 16GB limit might not fit, and they may have video or editing content that surpasses that.

- however for other organizations/people, they might actually want to limit to reasonable sizes that their work or policy allows, such as anywhere from 200MB to 5GB, or even wish to disallow big uploads all together, to a smaller 8MB-64MB to basically allow only typical documents and pictures, that are work related, and no piracy or large file storing to not overuse storage.

PHP by default sets upload limits to 2MB which would work for the majority of sites back in the early 2000s but not any more with current phone cameras - just to say that defaults do not always work and sometimes they need adaptation to particular needs.

I haven't commented the other changes:

- I don't understand the 3600s of max execution time and max input time that is being used by default. It is dangerous to let a script that is exposed to the internet or intranet run that long. I may be wrong but 3600s or longer is something I could expect from a cronjob, running in background in CLI, not via web browser, to process, convert, index, whatever files exist in a storage. I would not expect longer than anywhere from 120s to 600s for active web PHP processes in any application, and these are already high.

- hence thus I also implemented these config changes.

- But I think the PHP CLI limits also need config support as to avoid never ending cronjobs that will clog and kill the system if they accumulate. Easy to do math: every 15m, with cronjobs taking 3600s...

Changing Apache HTTPd config is also a bit awful for newcomers, however current default limits seem to be set to unlimited.

lpalgarvio:
Dunno if this snap building is documented anywhere, but I could not find it, so I'm experimenting and documenting here.

```
lxc launch ubuntu:16.04 nextcloud-snap
lxc exec nextcloud-snap -- bash
snap install snapcraft --classic
git clone https://github.com/lpalgarvio/nextcloud-snap.git
cd nextcloud-snap/snap
snapcraft
```

> This snapcraft project does not specify the base keyword, explicitly setting the base keyword enables the latest snapcraft features.
> Read more about bases at https://docs.snapcraft.io/t/base-snaps/11198
> Sorry, an error occurred in Snapcraft:
> Failed to pull source: unable to determine source type of 'src/redis/'.
> Check that the URL is correct or consider specifying `source-type` for this part. See `snapcraft help sources` for more information.

kesselb:
Your probably already know: Bigger files are uploaded in chunks if the Nextcloud clients are used.

Fuseteam:
Uh iiuc the mindset is that they shouldn't need to change anything, the largest should be allowed out of the box

kyrofa:
@Fuseteam is spot on. The fewer knobs the better. "Please let me upload larger files" shouldn't be a knob.

lpalgarvio:
> 
> 
> Uh iiuc the mindset is that they shouldn't need to change anything, the largest should be allowed out of the box

if defaults are set too low, people cant upload.
if defaults are set too high, or in the lines of "unlimited", creates security risks (uploading extremely large files to exhaust storage, uploading copyrighted files, generally abusing the systems)

the fact that apache and nginx web servers, and php (and php-fpm) support changing upload, http request body limits and such, memory limits and execution times - is there for a reason

jaded0:
I'm currently running into issues installing  Collabora Online - Built-in CODE Server as a direct result of this issue. It's worth reconsidering your decision to do this, since I would say that the simple installation of apps constitutes reason enough to allow for max_execution_time to be modified with the snap.

Fuseteam:
@jaded0 it is likely that there's a different cause, i have the CODE server installed fine

jaded0:
Yeah, I have extra slow internet, so when I finally found and ran a command to increase max_execution_time that's exactly what fixed it.

Fuseteam:
sounds fair, which command did you run and how long did you make the max_execution_time?

PR #1198, opened 2020-01-02
nextcloud: update to 16.0.7

kesselb:


stondino00:
Any chance you could build this with the updated 7.2 or 7.3 php for security fixes? That's what I'm waiting on.
https://www.php.net/ChangeLog-7.php#7.3.13
https://www.php.net/ChangeLog-7.php#7.2.26

stondino00:
Upgrading from 16.0.6 PR snap has no issues for me.

kyrofa:
I'm afraid the issue with 16.0.6 is still not fixed, here. https://github.com/nextcloud/server/pull/18259 doesn't seem to fix anything. If you install from stable (still 16.0.5):

    $ sudo snap install nextcloud

Create an account etc., then upgrade to 16.0.7 using this PR:

    $ sudo snap refresh nextcloud --channel=beta/pr-1198

Once it completes the upgrade you'll see this:

Screenshot from 2020-01-06 08-47-37

kesselb:
Unfortunately. I'm actually created this pr to do a snap build. Having issues to do it locally. I logged a issue for the problem and also found the code in question: https://github.com/nextcloud/server/issues/18673. Still need to figure out how to fix it properly. Shipping the latest `files_rightclick` app will workaround this behaviour. Actually this issue could also happen from 16.0.4 to 16.0.5. 

I know you are not a fan of pre- / after upgrade routines (and me neither). Enforcing a update for the files_rightclick app (occ app:update files_rightclick) and running occ upgrade afterwards will also prevent this issue.

It's probably not bad at all (if possible) to update all apps (occ app:update --all) to their latest available version before doing the actual upgrade. I think this groupfolder upgrade issue for some people is also related to an outdated groupfolder app.

kyrofa:
@kesselb you have been an amazing help, thank you for chasing that down.

> I know you are not a fan of pre- / after upgrade routines (and me neither). Enforcing a update for the files_rightclick app (occ app:update files_rightclick) and running occ upgrade afterwards will also prevent this issue.

Yeah, ignoring my gut "we shouldn't have to do that" reaction, my biggest problem with this sort of hack is that you can find yourself in the position of never being able to remove it!

> It's probably not bad at all (if possible) to update all apps (occ app:update --all) to their latest available version before doing the actual upgrade.

That seems less silly than updating just the rightclick app, and dare I say is perhaps behavior expected from something that automatically updates. @pachulo, what do you think?

pachulo:
> That seems less silly than updating just the rightclick app, and dare I say is perhaps behavior expected from something that automatically updates.

@kyrofa from my perspective that should only harm the use case where a user actively doesn't want to update a certain app, but I'm inclined to think that there shouldn't be many snap users doing it, if any....and besides that, it indeed looks like something that people could expect the snap to do: automatically update all the installed apps when updating nextcloud.

But anyway, I think that @kesselb & yourself know much more than I do regarding all the mess that app handling is, so I trust you on whatever decision you make to amend nextcloud update problems on the snap.

kesselb:
> automatically update all the installed apps when updating nextcloud.

That will happen as part of upgrade process. I'm suggesting to update all apps to their latest version before the actual upgrade. People are running outdated versions of apps. Sometimes app updates required to make the actual upgrade to a newer (nextcloud) version work: `occ app:update --all && occ upgrade` and then do the nextcloud upgrade.

In the long run that's something nextcloud upgrade process should do. Make all apps up to date, run update to next major, make all apps up to date again. A bit weird ;) It's possible that apps are compatible with more than one major (like files_rightclick). Other apps always stick to a specific major.

kyrofa:
Thanks you two, I'll test it out and verify that it solves the issue for v16 upgrades anyway, then we can continue focusing on getting to v17.

kesselb:
Thanks :+1: 

`pre-refresh` would be the hook right? I don't know much about snap package development. Just looking around a bit.

kyrofa:
These days, yes, although this snap predates the existence of that hook, so the upgrade stuff happens [here](https://github.com/nextcloud/nextcloud-snap/blob/master/src/nextcloud/bin/setup-nextcloud#L66).

kyrofa:
Interestingly, even on a fresh install, if one runs `occ app:update --all` Nextcloud immediately thinks it needs to upgrade and won't continue until you also run `occ upgrade`.

kyrofa:
Alright, https://github.com/nextcloud/nextcloud-snap/pull/1207 seems to fix this issue. It's ugly, and I'm not happy we need to do this, but at least it works.

kyrofa:
I've gone ahead and rebased this.

PR #1204, opened 2020-01-06
Release 16.0.5snap4

kyrofa:
- php: update to 7.3.13


kyrofa:
Tested both fresh install and upgrade, all seems to work well.

PR #1207, opened 2020-01-12
nextcloud-fixer: update all apps

kyrofa:
Nextcloud upgrades don't update apps properly (see nextcloud/server#18673). This results in a very poor upgrade experience if a bundled app needs to update. This PR works around this issue by updating apps automatically.

### DIFF
--- a/src/nextcloud/bin/nextcloud-fixer
+++ b/src/nextcloud/bin/nextcloud-fixer
@@ -24,6 +24,13 @@ if nextcloud_is_installed; then
 	trap 'disable_maintenance_mode' EXIT
 
 	occ -n db:convert-filecache-bigint
+
+	echo "Updating all apps..."
+	if occ -n app:update --all; then
+		# app:update downloads and extracts the updates, but now we
+		# need to run upgrade to run database migrations, etc.
+		occ -n upgrade
+	fi
 else
 	wait_for_nextcloud_to_be_installed
 
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -102,7 +102,7 @@ apps:
 
 hooks:
   configure:
-    plugs: [network, network-bind]
+    plugs: [network, network-bind, removable-media]
 
 parts:
   apache:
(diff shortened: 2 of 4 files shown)


src/nextcloud/bin/nextcloud-fixer
```diff
@@ -24,6 +24,12 @@ if nextcloud_is_installed; then
 	trap 'disable_maintenance_mode' EXIT
 
 	occ -n db:convert-filecache-bigint
+
+	echo "Updating all apps..."
+	if occ -n app:update --all; then
+		# No idea why, but Nextcloud wants an upgrade right after
+		occ -n upgrade
```
kyrofa: @kesselb any idea why this is required?

kyrofa:
You can pretty easily test this by installing nextcloud from stable:

    $ sudo snap install nextcloud

Noting that the rightclick app is out of date, and then refreshing to the snap from this PR:

    $ sudo snap refresh nextcloud --channel=beta/pr-1207

Once the refresh completes, you should notice that rightclick is now magically up to date.

kyrofa:
@pachulo I assume you've already updated the rightclick app, so I didn't actually bother building this for arm. Just sanity check the code, please.

src/nextcloud/bin/nextcloud-fixer
kesselb: Yes. occ app:update download the app updates and extract the code to the right place. Next step is to run occ upgrade routine (migrations, bump app version). I don't know why. Probably a "design" design in the past. The same process for updates via the official updater. Updater will fetch the latest release, put the code in place, run occ upgrade routine.

kesselb: After running snap refresh and occ upgrade all apps should be up to date (occ upgrade will check the appstore for updates). But https://github.com/nextcloud/server/blob/master/core/Command/App/Update.php will also return 0 if nothing happens.

kyrofa: Ah this is by design then. Seems a little odd, but that makes me feel better! It's okay if the update step returns zero, the upgrade step also returns zero if it's already up to date (despite a comment I have to the contrary, so I think that has been different in the past).

kyrofa: I've updated the comments here to reflect reality as I understand it, thank you for the explanation.

PR #1215, opened 2020-01-13
nextcloud-fixer: upgrade and update apps before anything else

### LINKED ISSUE #1214: Nextcloud installation broken after last refresh
Please, how do I debug the following problem? I am running a fairly important/big Nextcloud instance and googling turns up nothing:

I am on a ubuntu 18-04 server and this morning the Nextcloud installation on this server broke after an automatic refresh. 

When I go to the webpage of the installation I only get:

"Downgrading is not supported and is likely to cause unpredictable issues (from 16.0.7.1 to 16.0.5.1) "

I am using the stable version:

$ snap list nextcloud
Name       Version      Rev    Tracking  Publisher   Notes
nextcloud  16.0.5snap4  18019  stable    nextcloud✓  -

Any help very much appreciated.

kyrofa:
This PR fixes #1214 by making sure an upgrade and app update happens before anything else. It also refactors nextcloud-fixer to be a more easily-expanded set of fixes (hacks) by making them all shell scripts instead of lines in the same file.

### DIFF
--- a/src/nextcloud/bin/nextcloud-fixer
+++ b/src/nextcloud/bin/nextcloud-fixer
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/sh -e
 
 # shellcheck source=src/apache/utilities/apache-utilities
 . "$SNAP/utilities/apache-utilities"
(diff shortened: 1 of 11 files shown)


kyrofa:
Tested extensively on dev machine (fresh install and update) as well as update on production instance. This seems to work better than the last attempt.

PR #1221, opened 2020-01-16
nextcloud: notify admins upon update

### LINKED ISSUE #1220: Idea for the future: notify admins
As I stated in https://github.com/nextcloud/nextcloud-snap/issues/1129

>There is a working script that informs all admin users in the Nextcloud VM now. Maybe it could also get used here?

> The current version was merged [here](https://github.com/nextcloud/vm/pull/1030/files) but we are currently reworking the function to make it more efficient [here](https://github.com/nextcloud/vm/pull/1034/files).

kyrofa:
The main sell of the Nextcloud snap is that is updates automatically in a transactional manner. A side effect of this is that admins could completely miss the fact that an update occurred, and if the update did something they didn't expect they don't immediately know what to blame.

This PR resolves #1220 by adding the ability to notify all admins when the snap updates, making sure they know how to revert if something doesn't work properly.

### DIFF
--- a/src/common/utilities/common-utilities
+++ b/src/common/utilities/common-utilities
@@ -50,6 +50,16 @@ wait_for_command()
 	fi
 }
 
+get_previous_snap_version()
+{
+	snapctl get private.snap.version
+}
+
+set_previous_snap_version()
+{
+	snapctl set private.snap.version="$1"
+}
+
 enable_maintenance_mode()
 {
 	if run_command "Enabling maintenance mode" occ -n maintenance:mode --on; then
--- a/src/nextcloud/fixes/existing-install/99_notify_admins.sh
+++ b/src/nextcloud/fixes/existing-install/99_notify_admins.sh
@@ -1,0 +1,25 @@
+#!/bin/sh -e
+
+# shellcheck source=src/nextcloud/utilities/nextcloud-utilities
+. "$SNAP/utilities/nextcloud-utilities"
+# shellcheck source=src/common/utilities/common-utilities
+. "$SNAP/utilities/common-utilities"
+
+previous_version="$(get_previous_snap_version)"
+if [ "$previous_version" != "$SNAP_VERSION" ]; then
+	message="The Nextcloud snap updated itself to version $SNAP_VERSION. We"
+	message="$message are dedicated to ensuring these updates work amazingly"
+	message="$message well, but in the unlikely event something broke,"
+	message="$message remember you can revert the update with a single"
+	message="$message command:\n\n"
+	message="$message    $ sudo snap revert nextcloud\n\n"
+	message="$message Please also don't forget to log an issue:"
+	message="$message https://github.com/nextcloud/nextcloud-snap"
+
+	run_command \
+		"Notifying admins of update from ${previous_version:-unknown version} to $SNAP_VERSION" \
+		nextcloud_notify_admins \
+			"Nextcloud updated" "$(printf "%b" "$message")" || true
+
+	set_previous_snap_version "$SNAP_VERSION"
+fi
--- a/src/nextcloud/fixes/fresh-install/99_record_snap_version.sh
+++ b/src/nextcloud/fixes/fresh-install/99_record_snap_version.sh
@@ -1,0 +1,7 @@
+#!/bin/sh -e
+
+# shellcheck source=src/common/utilities/common-utilities
+. "$SNAP/utilities/common-utilities"
+
+# Record the snap version so we can notify when it's been updated
+set_previous_snap_version "$SNAP_VERSION"
--- a/src/nextcloud/utilities/nextcloud-utilities
+++ b/src/nextcloud/utilities/nextcloud-utilities
@@ -30,6 +30,22 @@ wait_for_nextcloud_to_be_installed()
 	wait_for_command "Waiting for Nextcloud to be installed" nextcloud_is_installed
 }
 
+# nextcloud_notify_admins <short message> <long message>
+nextcloud_notify_admins()
+{
+	if ! occ app:list --output=json | jq -e '.enabled | .notifications' > /dev/null; then
+		echo "Notifications app isn't enabled-- unable to send notification" >&2
+		return 1
+	fi
+
+	users=$(occ user:list --output=json | jq -r 'keys[]')
+	for user in $users; do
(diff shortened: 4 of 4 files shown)


the-sane:
:+1:

Question:  What's the cleanest way to handle a situation where a one-off notification may have to be sent to admins with specific instructions that only apply when upgrading from version X to version Y?  Should the `nextcloud-fixer` script just do a check against the current version and the next version, then run any of those types of things in a one-off function?

I'm imagining a worst-case scenario:  Things go horribly wrong with one update and the snap is in a situation where tomorrow the stable channel on version X will reach end of life, but pushing an update to version Y will break something and, whatever it is, it isn't going to be fixed in time.  It can't be automagically hacked around in a script so server admins will have to take action to fix it themselves after the update.  And then this scenario keeps happening so that, over time, multiple one-off sets of instructions have had to be pushed out as notifications.  Where do these one-off instructions live so they don't clutter things up too much?

What's the cleanest way to handle that?

the-sane:
Maybe the phrase one-off is a poor choice of words because you're right, those instructions would live on forever.  I meant that they'd only apply to one specific update scenario rather than being run for all updates.  I was trying to think of what might happen when there's one set of instructions for upgrading from X to Y, and another for X to Y+1, and another for Y to Z, etc etc.  Where do they live so things don't get incredibly cluttered?

Hopefully it is unlikely as you said, but I figure planning ahead now may save us some headaches later?

kyrofa:
> I meant that they'd only apply to one specific update scenario rather than being run for all updates.

That is actually what I'm saying is pretty unlikely. If they apply to the X->Y update, they will also mostly likely apply to the X->Y+1 update. Basically, any update where X is the starting point. That's semantics, though.

To really answer your question, yeah I suspect we could come up with something decent. We'd need something like this PR to track Nextcloud versions in the snap state so we know what we're upgrading from (it's possible `occ status` includes that, but I don't know off the top of my head). Then we just need to design a way for shell script to say "I need to run on an upgrade" and also have the ability to specify X as well as Y. It would probably look pretty similar to nextcloud-fixer and its set of shell scripts.

the-sane:
> Basically, any update where X is the starting point.

Ah, I see your point now.  I'm not sure why I was thinking otherwise because yes, that makes sense.

So nextcloud.occ status returns this:
```
  - installed: true
  - version: 16.0.7.1
  - versionstring: 16.0.7
  - edition: 
```
And it sounds like if that scenario ever happens, the current nextcloud-fixer setup could simply call a dedicated script containing all of these special per-version admin notifications.  So I'm satisfied that we won't have to re-engineer anything down the line to support that type of thing. :+1: 

As a side note, I've been looking through the scripts to familiarize myself with how they work.  I see that this admin notification script will get called by nextcloud-fixer, but I'm having trouble finding where nextcloud-fixer gets called in the first place during an upgrade?

kyrofa:
> As a side note, I've been looking through the scripts to familiarize myself with how they work. I see that this admin notification script will get called by nextcloud-fixer, but I'm having trouble finding where nextcloud-fixer gets called in the first place during an upgrade?

It's actually a service unto itself that runs once and exits (unless it exits non-zero, in which case it's run again): https://github.com/nextcloud/nextcloud-snap/blob/master/snap/snapcraft.yaml#L97

the-sane:
Aha, I see, thank you.

MarinMikael:
Hi, I've been following the discussions for few months now. I love reading dramas about updates rather than living them. X) Thank you for the great work !

About the usage of this script: 
Wouldn't it be interesting to add a dummy update before anything brakes just to send a notification a week before ? The message could even suggest them to use refresh.hold or refresh.timer to delay the update to whenever they want to do it. This way, sysadmins could do backups, test, and be there when they update, just in case.

About the-sane's remark on the script taking time if there is a lot of users: I would say that it matters even more in this case.

Also, if the process is really slow, maybe parallelizing would be interesting.
Here is a little example to save time if you'd like to try:
```
task() {
        echo hello $1; sleep 3; echo bye $1
}

L="0 1 2 3 4 5 6"
for i in $L; do
        task $i &
        pids[$i]=$!
done
for pid in $pids; do
        wait $pid
done

sleep 0.1; echo done
```

kyrofa:
> Hi, I've been following the discussions for few months now. I love reading dramas about updates rather than living them. X) Thank you for the great work !

Haha, best feedback we could get, thanks!

> Wouldn't it be interesting to add a dummy update before anything brakes just to send a notification a week before ? The message could even suggest them to use refresh.hold or refresh.timer to delay the update to whenever they want to do it. This way, sysadmins could do backups, test, and be there when they update, just in case.

It's not a bad idea, but I see a few potential issues:

1. Snaps are made to automatically update. Warning that an update is impending would seem to be information that isn't particularly interesting given that they installed the snap.
2. It seems somewhat scary to say "here's how you delay the update", thereby implying that the update might do something undesired. Contrast that with what is done in the PR, where we mention "FYI an update happened, here's what you do if it wasn't as good as it should be."
3. Given how the snap update process works, we can't be certain everyone actually got the warning update in a timely manner.
4. Most snap updates are security updates, where holding them for a week to give people a heads up sort of defeats the purpose of security updates.

> About the-sane's remark on the script taking time if there is a lot of users: I would say that it matters even more in this case.

This is only concerning if the notification is happening in a critical path somewhere. In this PR, for example, it's happening in a standalone service upon which nothing else depends. Nextcloud will fire up and be perfectly usable while the notifications are going out, even if it takes an hour. I don't think it's something we need to worry about; let's favor simplicity.

MarinMikael:
> 1., 2., 4. and 5.

I was thinking about the rare event where there is an update which is strongly suspected to break something, not something systematic.

> 4.

Alternatively, having a mailing list for important notifications would have the same purpose, with less effort while only targeting people who subscribed to it. An invitation to join this mailing list could be added to Nextcloud's snap install guide, which, may be the primary way though which sysadmins chose to use snap.

> Nextcloud will fire up and be perfectly usable while the notifications are going out

Sounds perfect.

kyrofa:
> I was thinking about the rare event where there is an update which is strongly suspected to break something, not something systematic.

Yeah we can probably make something happen as a special case. Not entirely reliable though, as I mentioned.

kyrofa:
Actually, @MarinMikael a good example of where what you're talking about is a GREAT idea is when we know we're releasing the final release of a track. For example, if someone is following 14/stable we can let them know that it's reached EOL and will no longer be updated.

PR #1223, opened 2020-01-19
CHANGELOG: update for 15.0.14snap1

kyrofa:
Also update README.


kyrofa:
Tested both fresh install and upgrade from 15/stable successfully.

PR #1244, opened 2020-02-09
hooks: add pre-refresh hook to update apps

### LINKED ISSUE #1243: Update apps before refreshing to next version
This will use the `pre-refresh` hook.

kyrofa:
Nextcloud's update process doesn't do a good job of ensuring apps are updated properly and re-enabled after the update. It seems to work best if apps are already as up-to-date as they can be. This PR resolves #1243 by taking a moment to make sure all apps are up-to-date if we know we're about to be refreshed.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -103,6 +103,8 @@ apps:
 hooks:
   configure:
     plugs: [network, network-bind, removable-media]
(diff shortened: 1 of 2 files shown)


kyrofa:
The snap builder won't produce a snap worth testing in this case, but I've testing this extensively locally.

PR #1250, opened 2020-02-13
snap: upgrade to core18

### LINKED ISSUE #1021: TLS 1.3 not supported
TLS 1.3 is enabled since Apache 2.4.36 (https://github.com/apache/httpd/blob/2.4.36/CHANGES), so it should be available on this snap ...

version : Version : 15.0.8.1

pachulo:
@kyrofa I known this won't be merged yet (context here https://github.com/nextcloud/nextcloud-snap/issues/755#issuecomment-583815600) but I thought that I would leave my work here and (try to) maintain it, so the changes are ready when we feel confident about the upgrade.

Meanwhile, it can also be tested by someone else than me.

Also, with the OpenSSL upgrade, we get TLS 1.3 without any changes on the Apache config, so this would also resolve #1021

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -53,7 +53,7 @@ apps:
     plugs: [network, network-bind]
 
   mysqldump:
-    command: mysqldump --defaults-file=$SNAP_DATA/mysql/root.ini --lock-tables nextcloud
+    command: run-mysqldump
     plugs: [network, network-bind]
 
   # Nextcloud occ command
@@ -207,19 +207,6 @@ parts:
 
       # Enable gmp
       - --with-gmp
-    stage-packages:
-      # These are only included here until the OS snap stabilizes
-      - libxml2
-      - libpng16-16
-      # These seems to be needed for core18
-      - libcurl4
-      - libjpeg9
-      - libbz2-1.0
-      - libmcrypt4
-      - libldap-2.4-2
-      - libfreetype6
-      - libgmp10
-      - libzip4
     build-packages:
       - libxml2-dev
       - libcurl4-openssl-dev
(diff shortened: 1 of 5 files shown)


snap/snapcraft.yaml
```diff
@@ -308,6 +318,9 @@ parts:
       # Remove scripts that we'll be replacing with our own
       - -support-files/mysql.server
       - -COPYING
+    stage-packages:
+      # This seems to be needed for core18
+      - libaio1
```
kyrofa: When specifying `base`, snapcraft no longer slurps up libraries from the build-packages to put into the snap. As a result, stage-packages are required. Rather than staging the entire deb though, can you stage only the libraries that are required (using the stage keyword, for example)?

pachulo: OK, I will try that, but I don't get what's the benefit of doing so, at least in this particular case, as the library is only 30KB (https://packages.ubuntu.com/bionic/libaio1) and it contains only a few files https://packages.ubuntu.com/bionic/amd64/libaio1/filelist ...what am I missing?

kyrofa: Indeed, this is general guidance. We'll end up with a lot of cruft we didn't have before if we just stage entire packages, and in some cases it will be a lot worse than 30KB. If it gets insane remember you can make stage-packages standalone parts as well, using the `nil` plugin.

kyrofa: It ended up being about 500K altogether. Not enough to justify the trouble, I think. Scratch my advice for now.

kyrofa:
Alright, I think we're ready for this. I pushed a follow-up commit to try and get this across the line, it seems to work locally without issue.

kyrofa:
Ah. This means the snap builder will no longer work. I'll try to budget some time to fix that soon, but I'll build this PR the old way for now.

kyrofa:
Alright, this is now in the `beta/pr-1250` channel for all architectures.

kyrofa:
Standard fresh install and upgrade test reveals nothing special about this. Everything works as expected. I'll put it on a staging server and give it a more thorough test soon.

kyrofa:
Yeah this seems to work great. Let's Encrypt integration still works, etc. Also:

Screenshot from 2020-06-05 11-50-02

TLS 1.3! I'm good with this as soon as we can get a non-amd64 tester in here. @pachulo you still slammed I suspect?

pachulo:
Did the upgrade on my arm64 instance and everything looks as good as before!

It's great to see this finally merged!

PR #1262, opened 2020-02-23
nextcloud: support notifying users with spaces in username

### LINKED ISSUE #1260: Admin notification doesn't support spaces in username
Today there was a bug report in the Nextcloud VM repository where the notify admin function didn't work. 
This was caused by spaces in the userid:
https://github.com/nextcloud/server/issues/18868

To fix that we are currently updating the function:
https://github.com/nextcloud/vm/pull/1105

Maybe you should think about improving the function as well?

kyrofa:
Nextcloud supports creating a user with spaces in the username, which confuses the update notification. This PR fixes #1260 by updating it to support spaces.

### DIFF
--- a/src/nextcloud/utilities/nextcloud-utilities
+++ b/src/nextcloud/utilities/nextcloud-utilities
@@ -38,8 +38,7 @@ nextcloud_notify_admins()
 		return 1
 	fi
 
-	users=$(occ user:list --output=json | jq -r 'keys[]')
-	for user in $users; do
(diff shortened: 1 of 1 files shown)


kyrofa:
Tested this successfully by installing from stable, creating an admin user with spaces in the username, and refreshing to the snap from this PR. Sends the notification without issue.

enoch85:
Thanks for this @kyrofa!

PR #1267, opened 2020-02-23
Disable v15 dailies

kyrofa:
v15 is no longer supported.


kyrofa:
Test ping @ros-security/approvers, does this work?

PR #1352, opened 2020-06-02
nextcloud: regenerate asset cache on startup

### LINKED ISSUE #1323: Nextcloud asset cache doesn't play nice with snap's revert capability
Hi,

I had a problem with my IMAP mailboxes just sitting 'Loading' but never loading in the last few days. After updating the ubuntu machine the problem was still there so I thought I would just try reverting to the previous snap installation. I issued the 'sudo snap revert nextcloud'  and it performed everything correctly. I rebooted the server but now when any user logs in via web browser the main page loads but none of the files. The IMAP still has the same problem so I am going backwards. Right now I would like to just get the files back! 

Notes
- If you use a nextcloud app such as that from an android phone then all files are there so there is nothing wrong with the files it is something to do with the browser access.

- Also it has an error regarding the cronjob not being run for the last 4 days! - this was not the case before the revert it was happy with the cronjobs.

Any suggestions to get me back to where we can see files via browser again?

kyrofa:
This PR resolves #1323 by regenerating the asset cache when the snap starts up. Without this, reverting the snap can result in an older Nextcloud attempting to use the assets of a newer Nextcloud, which is an experience that generally makes one sad.

### DIFF
--- a/src/nextcloud/fixes/existing-install/4_regenerate-assets.sh
+++ b/src/nextcloud/fixes/existing-install/4_regenerate-assets.sh
@@ -1,0 +1,7 @@
+#!/bin/sh -e
+
+# This does a lot of stuff, but what we really care about is the fact that it
+# regenerates the asset caches. This is required because these caches are
+# stored alongside the data, which is unversioned. Without clearing the caches
+# on starup, a revert would try to use a newer version's CSS, for example.
+occ -n maintenance:repair


kyrofa:
This works great. Install nextcloud from `stable` and refresh to `19/candidate/rc3` and then revert. Note all CSS is broken. Then uninstall Nextcloud and do the same experiment again, but instead of using `stable` use this PR's snap. The revert is beautiful! Thanks to @kesselb for the slick solution.

kesselb:
I'm happy to help. Keep the good work up :+1:

PR #1353, opened 2020-06-03
nextcloud: update to 19.0.0

### LINKED ISSUE #1342: Nextcloud: update to v19
Nextcloud 19 isn't released yet, but it's time to start testing its RCs at least.

kyrofa:
This PR resolves #1342 by updating Nextcloud to v19. It also enables the bcmath PHP module, and adds libfontconfig1 for collabora.


kyrofa:
So collabora doesn't appear to work. My logs are filled with:

> [05-Jun-2020 00:15:39 UTC] PHP Warning:  fsockopen(): unable to connect to localhost:9982 (Connection refused) in /var/snap/nextcloud/21308/nextcloud/extra-apps/richdocumentscode/proxy.php on line 186

Sounds like https://github.com/CollaboraOnline/richdocumentscode/issues/26 . @kesselb I shouldn't have to do anything special, right? It should Just Work?

szaimen:
Maybe not all system requirements are included in the snap?
https://github.com/CollaboraOnline/richdocumentscode/blob/master/README.md#system-requirements

kyrofa:
Quick update: making progress on that issue (which you can of course read for the details). Collabora devs are hoping to cut a fix for the current issue next week. Stay tuned.

kyrofa:
https://github.com/CollaboraOnline/richdocumentscode/issues/33 is resolved, just waiting for another collabora release with the fixes now.

kyrofa:
Alright, a new release has been cut. Can we get a few more tests of this now that Collabora works? Updated builds are available in the `beta/pr-1353` channel:

    $ sudo snap install nextcloud --channel=beta/pr-1353

    $ sudo snap refresh nextcloud --channel=beta/pr-1353

kyrofa:
I'm pleased to say that this seems to work well for me, both on a fresh install and on an upgrade (both with and without recommended apps installed in both situations).

mschwrdtnr:
Collabora isn't working for me with the Error Message "Could not establish connection to the Collabora Online server." 
It's possible that i made mistakes. I just refreshed the snap with `sudo snap install nextcloud --channel=19/edge` because with your command I get the error "error: requested a non-existing branch for snap "nextcloud": 19/beta/pr-1353" 

What's the problem?

kyrofa:
> What's the problem?

Ah, I keep forgetting about that. I updated the command above, can you try that again?

PnjDbq:
Ran the update from 18.0.6snap1
`nextcloud (beta/pr-1353) 18.0.6snap1+git3.c9f8006 from Nextcloud✓ refreshed
`
Gave the installer time to settle down and had two setup warnings:

```
Some files have not passed the integrity check. Further information on how to resolve this issue can be found in the documentation. (List of invalid files… / Rescan…)


The database is missing some optional columns. Due to the fact that adding columns on big tables could take some time they were not added automatically when they can be optional. By running "occ db:add-missing-columns" those missing columns could be added manually while the instance keeps running. Once the columns are added some features might improve responsiveness or usability.
Missing optional column "reference_id" in table "oc_comments".
```
Ignoring, as always and known, the imagick "warning".

Manually ran `sudo nextcloud.occ db:add-missing-columns` and that solved the optional columns issue.

I was able to get Collabora running (installed two apps as listed at https://nextcloud.com/blog/how-to-install-collabora-online-in-nextcloud-hub/).  Awesome that the snap supports this! 

Integrity check is still an open issue.  The list of impacted files is attached.
[Errors.txt](https://github.com/nextcloud/nextcloud-snap/files/4814557/Errors.txt)

PnjDbq:
> Collabora isn't working for me with the Error Message "Could not establish connection to the Collabora Online server." when using the Build-In-Code Server.

I had a similar strange issue.  The "use the default CODE" was already checked and throwing an error.  I had to uncheck it (by picking "use your own server" and recheck "use the built in CODE".
Annotation 2020-06-22 120133

mschwrdtnr:
@kyrofa it works for me, thanks!
The performance is very bad compared to the demo server. Is this an issue of the performance of the server or is it not really recommend to use built in server with the snap? My machine have 2 vCPUs and 16 GB RAM.

Got the same "Problem" like @PnjDbq .

kyrofa:
> Manually ran sudo nextcloud.occ db:add-missing-columns and that solved the optional columns issue.

Interesting, I wonder how old your installation needs to be to see this, but we can easily do this on the behalf of users (we already do so with missing indices). I'll look into it.

> Integrity check is still an open issue. The list of impacted files is attached.

I have no clue what this is. That list of files doesn't look like it relates to Nextcloud at all!

> The performance is very bad compared to the demo server. Is this an issue of the performance of the server or is it not really recommend to use built in server with the snap?

I honestly don't know enough about collabora to tell you one way or the other.

> Got the same "Problem" like @PnjDbq .

You mean the weird issue with "use the built in CODE"?

PnjDbq:
> > Integrity check is still an open issue. The list of impacted files is attached.
> 
> I have no clue what this is. That list of files doesn't look like it relates to Nextcloud at all!

There's Nextcloud documentation linked to the error:  
https://docs.nextcloud.com/server/19/go.php?to=admin-code-integrity

Super strange.  All the errors are "EXTRA_FILE"'s belonging to the core package.  Basically the exact thing the snap would manage.

Anyhow, rolled back and tried the update again.  `sudo snap revert nextcloud `to get back to 18.0.6snap1 and the same `sudo snap refresh nextcloud --channel=latest/beta/pr-1353` to upgrade.   No problems after reverting to v18.  Same two errors when refreshing to pr-1353:

Annotation 2020-06-22 155851

kyrofa:
Oh wow, I can actually duplicate this, I totally missed it. Investigating now.

kyrofa:
Oh that's embarrassing, this is due to the stage-package being added here. Super easy fix, but I would have missed it, thank you @PnjDbq. I'll ask you to confirm as soon as I get an updated build out.

mjdubbs:
I upgraded from 18.0.6snap1, then installed Collabora, then installed CODE sever.
I got the same errors as @PnjDbq but I didn't have to toggle "Use builtin CODE server"
Performance of Collabora seems good

kyrofa:
So I pushed an update that fixes all the problems discovered by @PnjDbq, but then Collabora broke: https://github.com/CollaboraOnline/richdocumentscode/issues/39 . @PnjDbq can you confirm that the integrity check and optional columns are no longer an issue?

PnjDbq:
Good news / bad news:

Integrity check and optional columns are resolved after the update.  Not seeing any issues with Collabora.

Bad news is now I’m seeing an issue with "External storages".  Not sure if this was in the version yesterday (probably was); just noticed it today.
Annotation 2020-06-23 142049

Getting 503 errors in both the desktop sync & web interface.  Tried a bunch of things and then came across https://github.com/nextcloud/server/issues/21429 and https://github.com/nextcloud/server/issues/20979.

Looks like the fix might be in 19.0.1?  Since my only reason for using SFTP to localhost is to escape snap confinement (vs the "local" method), are there other tricks that mitigate mount local data that's outside / media?

kyrofa:
Oh interesting, you're right, looks like the fix is in 19.0.1.

> are there other tricks that mitigate mount local data that's outside / media?

I actually wasn't aware of the trick you're using. Typically I recommend folks bind-mount into /media.

kyrofa:
(FYI, the collabora issue seems to have been resolved).

kyrofa:
Now that collabora is working and the integrity check and optional columns are taken care of, this seems to work for the common case, and generally that's what I use to judge what are blocking issues. I'm not going to push breakage at you though, @PnjDbq, especially after you took the time to give this a thorough few rounds of testing. I see a few options:

- You switch to `18/stable` until 19.0.1 is out, then switch back to `stable` on your own time
- You switch to using bind-mounts into /media as the confinement workaround instead of using external storage
- We hold off on pushing v19 out until 19.0.1

Which path would you prefer?

kyrofa:
By the way. Given that https://github.com/nextcloud/server/pull/21203 landed in stable19 a while ago, you should try out `19/edge` and confirm this issue is indeed not present there. Not a solution, I don't recommend staying there, but would make us sure that 19.0.1 will actually have the fix.

PnjDbq:
Please don't hold up for me.  I'm good with the solutions you mentioned.

>  you should try out `19/edge` and confirm this issue is indeed not present there. 

Great idea!   I gave that a try and the SFTP external storage issue is resolved with 19/edge.  The fix works and it's on the way...

kyrofa:
> Please don't hold up for me. I'm good with the solutions you mentioned.

Okay, I'm going to start the process then.

> The fix works and it's on the way

So glad to hear!

PR #1377, opened 2020-06-09
snap: rewrite plugins to workaround snapcraft breakage

kyrofa:
This PR is a backport of #1376 for v17.


kyrofa:
Depends on #1373 before tests will pass. Man that is some breakage. Landing in the red to get #1373 green.

PR #1472, opened 2020-10-02
Release 19.0.3snap2

kyrofa:
- php: disable output buffering
- php: update to 7.3.23
- php: include support for argon2


kyrofa:
Seems to work great, both as a fresh install and as an upgrade from stable.

PR #1511, opened 2020-10-19
migrations: support daily versions

### LINKED ISSUE #1505: snap refresh fails due to illegal version number
I am on Nextcloud 18/edge at the moment.
when I run 

    sudo snap refresh

I get the following error

> error: cannot perform the following tasks:
> Run post-refresh hook of "nextcloud" snap if present (run hook "post-refresh": /snap/nextcloud/23790/bin/run-snap-migrations: 48: /snap/nextcloud/23790/bin/run-snap-migrations: Illegal number: 18-2020-10-15)

I guess it is related to the recent changes in run-snap-migrations 

when I run  

    snap list

i get the following results:

> Name       Version        Rev    Tracking       Publisher   Notes
> core       16-2.47        10126  latest/stable  canonical✓  core
> core18     20200724       1885   latest/stable  canonical✓  base
> nextcloud  18-2020-10-12  23683  18/edge        nextcloud✓  -

so it seems that lines 44 to 48 in https://github.com/nextcloud/nextcloud-snap/blob/master/src/migrations/bin/run-snap-migrations have some expectations which are not met (at least by my installation).

Thanks in advance for looking into this.

kyrofa:
The current migration version check assumes a version of the format major.minor.patch, which isn't the case when it comes to daily builds. This PR fixes #1505 by updating the logic to take that into account, and bailing the version check entirely if major versions can't be reliably extracted.

### DIFF
--- a/src/migrations/bin/run-snap-migrations
+++ b/src/migrations/bin/run-snap-migrations
@@ -14,7 +14,12 @@ version_less_than()
 
 major_version()
 {
-	echo "$1" | sed -r 's/([0-9]+)\..*/\1/'
+	echo "$1" | sed -r 's/([0-9]+)[.-].*/\1/'
+}
+
+is_integer()
+{
+	expr "$1" : '^[0-9]\+$' > /dev/null
 }
 
 # Before we do any migrations, Nextcloud is very strict about what makes for a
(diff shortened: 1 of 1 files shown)


janschreier:
thanks for fixing this 👍

PR #1584, opened 2020-12-15
Document how to reduce hashing costs

andybalaam:
I have found that my Nextcloud snap installation works much better now I have reduced the hashing cost configuration.

Do you think it's worth documenting this here, or should I suggest it to upstream Nextcloud?

I am not certain of the security implications and couldn't find much information about it.  If we can find information, I'm sure the wording could be improved.


andybalaam:
Sorry, wrong branch.

Issue #1628, opened 2021-02-02
Collabora Online (richdocumentscode as part of Nextcloud Snap) should use system fonts


timar:
#### Describe the bug
Collabora Online renders documents on the server side. It uses bundled fonts and system fonts. From snap, system fonts are not available.

It is a problem, because the size of richdocumentscode is already big, we do not want to add more fonts to it. We aimed for reasonably good Unicode coverage with bundled Noto fonts, but customers, especially in Asia want more, and their needs could be better served with system fonts. 

For the success of snapped Nextcloud + Collabora Online on these markets, we need a way for the users to add fonts easily to their setup.

FynnFreyer:
Apparently snapd provides snaps with access to system fonts, so this issue might be worth to revisit/reopen

https://forum.snapcraft.io/t/desktop-allow-access-to-host-system-fonts/1796

PR #1649, opened 2021-02-14
Allow access to fonts on host (fixes: #1628)

timar:


kyrofa:
Sorry for the delay, I need to see if this interface is available on Ubuntu Core before I give it more thought and I just haven't had a chance yet.

PR #1700, opened 2021-04-18
POC how to remove index.php from the URL

szaimen:
This is a POC how the index.php can get removed. It is based on https://github.com/nextcloud/nextcloud-snap/commit/a3a324dff626e96ce1d45cad334a71751a36b095

---

It works this way: every push that is pushed to any branch that is not called master, develop or '[12][0-9]' (this regex covers 10-29) will trigger the new workflow that downloads the current branch, runs the script that I wrote (My script basically extracts all for the .htaccess file needed lines from the Setup.php file and writes it into my new .htaccess file.), changes my new .htaccess file to be based on the in the branch used nextcloud archive, commits the changes (if any) and pushes them to the same branch. This way you will be able to track any possible changes in the htaccess file in new Nextcloud versions in the future before merging into any of the mentioned branches.

If the feature gets enabled by the user it just includes my new .htaccess file in the apache conf which will automatically enable pretty urls for the users Nexcloud instance.

Unfortunately, I was only able to enable this feature when Nextcloud is runing on its own (sub)domain or ip-address and not in a subdirectory but I am fine with this shortcoming and this was already added to the readme file as disclaimer. 
BTW: maybe the snap is not intended to work in subdirectories, anyway? 
(I ask because I didn't manage to put the nextcloud instance into a subdirectory at all...)


szaimen:
@kyrofa would be great to get some early feedback on this :)

src/apache/conf/httpd.conf
```diff
@@ -193,3 +193,37 @@ ServerSignature Off
...
+    RewriteCond %{REQUEST_URI} !^/\.well-known/(acme-challenge|pki-validation)/.*
+    RewriteCond %{REQUEST_FILENAME} !/richdocumentscode(_arm64)?/proxy.php$
+    RewriteRule . index.php [PT,E=PATH_INFO:]
+    RewriteBase ${REWRITEBASE}
+    <IfModule mod_env.c>
+        SetEnv front_controller_active true
+        <IfModule mod_dir.c>
+        DirectorySlash off
+        </IfModule>
+    </IfModule>
```
szaimen, line 228: @kyrofa Do you think it is possible to automatically generate this from the Nextcloud tarball?
Maybe we could use some features in the snapcraft.yaml file or how could this work?

szaimen:
Unfortunately I cannot find out why circleci is failing. What is the reason?

src/apache/conf/httpd.conf
szaimen: I mean I was able to write a script that does that (https://github.com/nextcloud/nextcloud-snap/issues/412#issuecomment-789663913) but is it possible to execute it from the snapcraft.yaml?

src/nextcloud/utilities/nextcloud-utilities
```diff
@@ -71,3 +71,57 @@ set_previous_cronjob_interval()
 {
 	snapctl set private.nextcloud.cron-interval="$1"
 }
+
+# Based on https://docs.nextcloud.com/server/21/admin_manual/installation/source_installation.html?highlight=htaccess%20rewritebase#pretty-urls
+is_htaccess_rewritebase_valid()
+{
+	local REWRITEBASE
```
kyrofa, line 78: This isn't valid posix shell, looks like bash. That's why circle is failing-- this is using a posix shell.

szaimen: hm... I've removed that. But there seems to be something else that isn't compliant. Do you know what it is?

szaimen: I would resolve it myself but I cannot access the log that should be provided by circleci...

pachulo: This is the error that the static-tests CircleCI is reporting:
```
In /root/nextcloud-snap/tests/../src/nextcloud/utilities/nextcloud-utilities line 79:
	if ! [ "$REWRITEBASE" = '/' ] && ! [[ "$REWRITEBASE" =~ ^/[a-z0-9]+$ ]]; then
                                           ^-- SC2039: In POSIX sh, [[ ]] is undefined.
```

README.md
```diff
@@ -123,6 +123,17 @@ If you want to disable the cronjob completely, run:
 
 To reenable it again simply set the `nextcloud.cron-interval` snap variable to a value that isn't `-1`
 
+#### Nextcloud Prettylink configuration
+
+Before enabling this feature, please first set the correct `overwrite.cli.url` and `htaccess.RewriteBase` in Nextcloud by following [this doc](https://docs.nextcloud.com/server/latest/admin_manual/installation/source_installation.html#pretty-urls)
+
+Afterwards, if desired the snap can remove the index.php from the URL:
+
+    $ sudo snap set nextcloud nextcloud.prettylink=true
```
pachulo, line 132: Why don't you call the setting `nextcloud.prettyurls`?

szaimen: thats a good idea! I'll do that 👍

src/nextcloud/utilities/nextcloud-utilities
szaimen: thanks! I'll have a look how to fix that.

README.md
szaimen: done with 5bad0f0

src/nextcloud/utilities/nextcloud-utilities
szaimen: I think I fixed it with c33ae03

szaimen: Seems like this didn't work. What is now the problem?

pachulo: Now the static-tests have passed, so the problem is with the dynamic tests; all of them are failing, I guess because `httpd` is not starting properly:
```
Change operating mode
  production (FAILED - 1)
  debug (FAILED - 2)
  invalid
  
An error occurred in an `after(:context)` hook.
Failure/Error: fail "Timed out trying to access Nextcloud: #{uri.to_s}"

RuntimeError:
  Timed out trying to access Nextcloud: http://localhost
# ./spec/spec_helper.rb:219:in `rescue in wait_for_nextcloud'
# ./spec/spec_helper.rb:201:in `wait_for_nextcloud'
# ./spec/change_mode_spec.rb:6:in `block (2 levels) in <top (required)>'
# ------------------
# --- Caused by: ---
# Timeout::Error:
#   execution expired
#   ./spec/spec_helper.rb:215:in `sleep'

Change PHP memory limit
  unlimited (FAILED - 3)
  bytes (FAILED - 4)
  invalid (FAILED - 5)
  
An error occurred in an `after(:context)` hook.
Failure/Error: fail "Timed out trying to access Nextcloud: #{uri.to_s}"

RuntimeError:
  Timed out trying to access Nextcloud: http://localhost
# ./spec/spec_helper.rb:219:in `rescue in wait_for_nextcloud'
# ./spec/spec_helper.rb:201:in `wait_for_nextcloud'
# ./spec/change_php_memory_limit_spec.rb:4:in `block (2 levels) in <top (required)>'
# ------------------
# --- Caused by: ---
# Timeout::Error:
#   execution expired
#   ./spec/spec_helper.rb:215:in `sleep'

Change ports
  http
    http (FAILED - 6)
    
An error occurred in an `after(:context)` hook.
Failure/Error: fail "Timed out trying to access Nextcloud: #{uri.to_s}"

RuntimeError:
  Timed out trying to access Nextcloud: http://localhost
# ./spec/spec_helper.rb:219:in `rescue in wait_for_nextcloud'
# ./spec/spec_helper.rb:201:in `wait_for_nextcloud'
# ./spec/change_port_spec.rb:5:in `block (3 levels) in <top (required)>'
# ------------------
# --- Caused by: ---
# Timeout::Error:
#   execution expired
#   ./spec/spec_helper.rb:215:in `sleep'
  https
    https (FAILED - 7)
    http still redirects to unchanged https (FAILED - 8)
    http redirects to changed https (FAILED - 9)
    Let's Encrypt challenge request (FAILED - 10)
...
```

README.md
pachulo: Again, why not `nextcloud.prettyurls` instead  of `nextcloud.prettyurl`?

szaimen: sry, I haven't read accurate enough. should be done with 1e2a99a

src/nextcloud/utilities/nextcloud-utilities
szaimen: I think I've figured out what the problem was. Should be fixed with da2814e

src/apache/conf/httpd.conf
```diff
@@ -193,3 +193,38 @@ ServerSignature Off
 <IfDefine EnableHTTPS>
     Include ${SNAP}/conf/ssl.conf
 </IfDefine>
+
+<IfDefine EnablePrettyurls>
+    # From here: https://github.com/nextcloud/server/blob/stable21/lib/private/Setup.php#L523-L561
```
pachulo, line 198: Shouldn't we be using the `stable20` version?

szaimen: I thought thid would make sense to implement with NC21? Or do you wanna backport it?

pachulo: Well, there's still no ETA for the snap using nextcloud 21, so I'm not sure what will come first: this functionality or that

szaimen: Okay, I mean as I've stated at the very top of this issue is that probably extracting the specific lines of code out of the nextcloud tarball would probably the best way of doing it . But I don't know if that is posibble with snaps...

pachulo: Yes, I think that it can be doable, I just don't know exactly how (maybe you can ask for help on https://forum.snapcraft.io/)...meanwhile, to test if this works, I think that we should adapt it to the current latest version nextcloud on the snap, 20 in this case

szaimen: I've used the config now from stable20 but it seems like nothing changed concerning the config between those two versions.

szaimen: I think maybe it makes sense to not run the script from the snapcract.yml but maybe via github actions. This would allow to see any change to the file on new PRs.

szaimen: I'll try to figure out how to do that.

szaimen: should be done

src/apache/conf/httpd.conf
szaimen: I think I've found a way how to do this

szaimen:
@pachulo I think circleci is failing because my new script is written in bash. Is there a way to exclude my script from getting scanned by circleci or shall I try to remove all incompatible things?

.github/workflows/get-prettyurl-conf.yml
```diff
@@ -0,0 +1,29 @@
...
+          ./prettyurl.sh
+      - name: Commit files
+        run: |
+          git config --local user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git config --local user.name "github-actions[bot]"
+          git commit -m "Add prettyurl conf" -a
+      - name: Push changes
+        uses: ad-m/github-push-action@master
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
szaimen, line 28: to make this new workflow work, someone with configure rights on the repository needs to add a GITHUB_TOKEN secret. See https://docs.github.com/en/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-a-repository. It needs to be called GITHUB_TOKEN. Otherwise I could change its name.
@kyrofa Could you do this?

szaimen: it seems like this has worked also without adding the token

szaimen:
I think this should be ready :)

szaimen:
Now I am satisfied :)

szaimen:
Now that all checks are green, how can we test if it actually works?

pachulo:
I guess you should be able to build the snap and test this yourself. Another option is to setup a `snap package` on Launchpad and have it build for you there. More info: https://www.youtube.com/watch?v=X_U-pcvBFrU

szaimen:
I've tested this successfully locally. The only question that remains is why it disables itself during a restart of the snap automatically. What am I missing?

src/apache/bin/httpd-wrapper
```diff
@@ -25,6 +27,19 @@ else
 	echo "No certificates are active: using HTTP only"
 fi
 
+if should_enable_prettyurls; then
+    if is_htaccess_rewritebase_valid; then
```
szaimen, line 31: is maybe this the problem during a restart?
Can the function retreive any value from nextcloud during a restart?

szaimen: seems like this was the case. is fixed now.

szaimen:
I've tested this PR successfully including restarts of the snap. Would love to get some feedback now :)

szaimen:
Sry, I've just seen that my change to the htaccess file results in an integrity error.
So this must be done a bit differently.

szaimen:
This should now finally work. I've successfully worked around the integrity check 👍

kyrofa:
Nice work here @szaimen, I've got a few ideas and I still need to give it a more thorough review, but the fact that you got this working at all is wonderful!

pachulo:
Great work @szaimen ! Just one thing, could you edit your first comment here with the details of solution that you implemented in the end?

szaimen:
> Just one thing, could you edit your first comment here with the details of solution that you implemented in the end?

I've just wrote it into the description of the PR. Hope that is okay, too?

prettyurl.sh
```diff
@@ -0,0 +1,55 @@
...
+# git clone --branch prettyurl https://github.com/nextcloud/nextcloud-snap.git
+# cd nextcloud-snap
+
+# Get the correct tar file from the branch itself
+NC_DOMAIN="$(grep 'https://download.nextcloud.com/server' ./snap/snapcraft.yaml | grep -oP 'https://.*')"
+
+# Download the tar file
+curl -fL "$NC_DOMAIN" -o /tmp/ncpackage.tar.bz2
+
+# Extract the archive
```
uPvtkBlpiucvDUf, line 14: I think the sha256 hash of 'ncpackage.tar.bz2' should be checked against the one listed in snapcraft.yml file here.

szaimen: I don't think that this is necessary because it only uploads any change to the .htaccess file jn the curent origin developing branch and not any production branch. But I'll let @kyrofa decide!

szaimen:
@kyrofa any update here? I'd like to push this a bit forward :)

kyrofa:
Not yet, I'm trying to finish off the mysql update first, hang tight!

.github/workflows/get-prettyurl-conf.yml
```diff
@@ -0,0 +1,34 @@
+name: get prettyurl conf
```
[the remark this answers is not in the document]
szaimen: The bot-authored commits are not random. They only get commited if anything in the new htaccess file changed because the nextcloud package was upgraded

prettyurl.sh
```diff
@@ -0,0 +1,55 @@
...
+then
+    echo "The Setup couldn't get extracted."
+    exit 1
+fi
+
+# Alway create a new htaccess file
+rm -f ./src/apache/conf/.htaccess
+
+# Get content of the updateHtaccess function and store it in a Variable called updateHtaccess
+updateHtaccess="$(sed -n "/function updateHtaccess/,/function/p" /tmp/nc/nextcloud/lib/private/Setup.php)"
```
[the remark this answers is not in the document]
szaimen: Because of the mentioned fragility I've come up with this solution so that you will be able to monitor any possible change in the file before you merge it into develop or any stable branch. 
@kyrofa I don't think that upstream will be able to to move all of that into a template file that we can use because they need to modify the template based on your webserver settings and need to append it to the alread existing htaccess file.
So if you should decide that you don't like my aproach, we can probably close this PR now and not in the future since there will be no other way!
But lets see what @kesselb says.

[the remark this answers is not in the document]
kyrofa: Hmm, we might also be able to just run the PHP command that utilizes this instead, have you experimented with that at all?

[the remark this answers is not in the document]
szaimen: I guess the problem is that the needed function which inserts the lines into the htaccess won't  work without deploying a complete Nextcloud, as far as I understand it and this is honestly nothing I want to look into. To make it work without a Nextcloud instance, we would need to modify the script, too. That's why I thought that modifying the file directly without needing to set up the php stack, etc. would make more sense...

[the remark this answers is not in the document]
szaimen: And since we can monitor the new htaccess file this way and have confirmed that it currently works, I thought it was feasible and maintainable this way...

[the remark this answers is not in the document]
szaimen: Ping @kyrofa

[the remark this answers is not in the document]
tawissus: I don't know if it helps but if I remember correctly Nextcloud Pi has this option as a setting. Maybe you can find something there in the code?

https://ownyourbits.com/nextcloudpi/

https://docs.nextcloudpi.com/en/configuration-reference/#nc-prettyurl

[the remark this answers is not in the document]
szaimen: No, it doesn't help unfortunately because the snaps .htaccess is read only.

szaimen:
I'll open a new one to clean the history

szaimen:
Superseded by https://github.com/nextcloud/nextcloud-snap/pull/1794

PR #1715, opened 2021-04-28
Add permission-policy header, opt out of FloC

ph00lt0:
Google is trying to enlarge it's monopoly with FloC. We should put all our efforts in blocking Google from doing so. 

Recently WordPress and Github decided to block FloC already, Nextcloud cannot be left behind, hence my proposal. Feel free to relocate this, if you believe this header should be placed elsewhere. 

More info on FloC:
https://www.eff.org/deeplinks/2021/03/google-testing-its-controversial-new-ad-targeting-tech-millions-browsers-heres

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+v 20.0.9snap1
+  - mysql: update to 5.7.34
+  - nextcloud: update to 20.0.9
+
 v 20.0.8snap1
   - logrotate: rotate apache access log
   - redis: update to 5.0.12
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 Nextcloud server packaged as a snap. It consists of:
 
-- Nextcloud 20.0.8
+- Nextcloud 20.0.9
 - Apache 2.4
 - PHP 7.3
 - MySQL 5.7
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -345,7 +345,7 @@ parts:
   mysql:
     plugin: cmake
     source: https://github.com/mysql/mysql-server.git
-    source-tag: mysql-5.7.33
+    source-tag: mysql-5.7.34
     source-depth: 1
     override-pull: |
       snapcraftctl pull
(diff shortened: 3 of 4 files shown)


ph00lt0:
@kyrofa any info on how to get this DCO correct? can't seem to figure out. Haven't used this before.

kyrofa:
As I understand it, Google will [default to using every website that serves ads](https://github.com/WICG/floc/issues/45#issuecomment-781042491), which does not include the default (and probably average) Nextcloud instance. Also, I'm not really interested in adding random headers for random services-- if you don't want to be sold, stop using browsers made by people who _make a living selling you_.

ph00lt0:
Personally I obviously do not use chrome. But you don't know what the receiving party of shared files uses. Unfortunately a huge chunk of the internet users is using Chrome. From my understanding of FloC all websites will be somehow collected, but I might be wrong. And yes I agree adding headers isn't ideal but seems we have to do this nowadays for all sorts of security settings that should be default...

> As I understand it, Google will [default to using every website that serves ads](https://github.com/WICG/floc/issues/45#issuecomment-781042491), which does not include the default (and probably average) Nextcloud instance. Also, I'm not really interested in adding random headers for random services-- if you don't want to be sold, stop using browsers made by people who _make a living selling you_.

JanOstrowka:
“alternative identifiers to track individuals as they browse across the web.”
- Google, source: [theverge.com](https://www.theverge.com/2021/3/30/22358287/privacy-ads-google-chrome-floc-cookies-cookiepocalypse-finger-printing)

But why not include it either way? It's not going to make any additional harm

ph00lt0:
@kyrofa the comment you are referring to is talking about the "Origin Trial", whatever that means. From my understanding the final version will collect all websites browsed by chrome users. I might be wrong tho, but this is what I am been told.

kyrofa:
Sorry @ph00lt0, that came off as grumpy, and rightly so, but definitely not toward you. I appreciate your efforts here, but I refuse to work around Google's latest evilness when, in reality, the majority of the internet is okay with being sold. For the few of us that aren't, we don't use chrome, and this doesn't effect us. The fix is all client-side, as they say.

> And yes I agree adding headers isn't ideal but seems we have to do this nowadays for all sorts of security settings that should be default...

Things that are at least de-facto standards, like HSTS, feel different to me. Adding headers to opt out of this or that service will never end.

PR #1723, opened 2021-05-09
Readme.md - Show correct php version

szaimen:
Signed-off-by: szaimen <szaimen@e.mail.de>


kyrofa (approved):
Ah, thank you. Note that we don't land PRs straight to master generally, so I changed the PR base, but looks good to me!

PR #1730, opened 2021-05-16
nextcloud: update to 21.0.1

### LINKED ISSUE #1727: nextcloud: update to v21

kyrofa:
This PR resolves #1727 by upgrading Nextcloud to the latest v21 release.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -188,8 +188,8 @@ parts:
 
   nextcloud:
     plugin: dump
-    source: https://download.nextcloud.com/server/releases/nextcloud-20.0.9.tar.bz2
-    source-checksum: sha256/c8fe4ae86bc51276be6d39cbaf3f45f3b29df901cd5a4163b8a6af0e09e9ff04
+    source: https://download.nextcloud.com/server/releases/nextcloud-21.0.1.tar.bz2
+    source-checksum: sha256/dd7c8ccc01547914a75b44bbf86028289c8919dc39f4e2e720147b6bd596aebe
     organize:
       '*': htdocs/
       '.htaccess': htdocs/.htaccess


kyrofa:
Builds of this PR for all architectures is available in the `beta/pr-1730` branch. Please help test!

fede1581:
nextcloud.occ status shows:
The current PHP memory limit is below the recommended value of 512MB.
  - installed: true
  - version: 21.0.1.1
  - versionstring: 21.0.1
  - edition: 

My server's Overview also shows version 21.0.1. And some new errors:
- A background job is pending that checks for user imported SSL certificates. Please check back later.
- Your installation has no default phone region set. This is required to validate phone numbers in the profile settings without a country code. To allow numbers without a country code, please add "default_phone_region" with the respective ISO 3166-1 code ↗ of the region to your config file.

https://scan.nextcloud.com/ says that i'm running Nextcloud 20.0.4.0.

Despite that things, everything seems to work fine

Pilzinsel64:
Hello.

I just tested tested this build with my private instance.
Seems to work fine and already very fast. Thank you very much for the work! 👌

The only points I regoniced are ...
- I had to run "nextcloud.occ upgrade" and turn off the maintenance mode manually (maybe my fault by just not waiting long enough?)
- Whiteboard Integration can be installed but it seems that the process can not be started. When I try to open or create a whiteboard file, the following error comes up. (Should I open an issue at the whiteboard integration repo?)

Not formatted:
```
{"reqId":"FycOeiRwSJZSo46tNTbO","level":2,"time":"2021-05-17T12:55:27+00:00","remoteAddr":"5.100.yyy.xxx","user":"zzzzzz","app":"integration_whiteboard","method":"GET","url":"/index.php/apps/integration_whiteboard/space/3330585","message":"Spacedeck request connection error : cURL error 7: Failed to connect to localhost port 9666: Connection refused (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for http://localhost:9666/api/spaces","userAgent":"Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:88.0) Gecko/20100101 Firefox/88.0","version":"21.0.1.1","id":"60a27ff0b5bca"}
```
Formatted:
```
[integration_whiteboard] Warning: Spacedeck request connection error : cURL error 7: Failed to connect to localhost port 9666: Connection refused (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for http://localhost:9666/api/spaces

GET /index.php/apps/integration_whiteboard/space/3330585
from 5.100.yy.xxx by zzzzzz at 2021-05-17T12:55:27+00:00
```

kyrofa:
> The current PHP memory limit is below the recommended value of 512MB.

That's [easy to change](https://github.com/nextcloud/nextcloud-snap#php-memory-limit-configuration) for yourself, if you so desire.

> Your installation has no default phone region set.

Okay that's not really an error, just a suggested configuration setting. The snap won't be able to pick a reasonable default there, so I guess it makes sense. Thanks for testing, @fede1581!

> I had to run "nextcloud.occ upgrade" and turn off the maintenance mode manually (maybe my fault by just not waiting long enough?)

Hmm, that's hard to know, @Pilzinsel64. There is indeed a job called `nextcloud-fixer` that runs a number of things , and it will enable maintenance mode (and disable it after it's done), so my gut reaction is that you accessed the web interface right between when it finished upgrading and before the fixer flipped on maintenance mode. Remember you can always run `sudo snap logs -f nextcloud` to get an idea of whether it's doing anything before forcing it one way or another.

Pilzinsel64:
> Hmm, that's hard to know, @Pilzinsel64. There is indeed a job called nextcloud-fixer that runs a number of things , and it will enable maintenance mode (and disable it after it's done), so my gut reaction is that you accessed the web interface right between when it finished upgrading and before the fixer flipped on maintenance mode. Remember you can always run sudo snap logs -f nextcloud to get an idea of whether it's doing anything before forcing it one way or another.

Yes, probably I just tried to access the web interface to early. Don't really know how long it would take.

> Yeah I'd start there, I know nothing about this app. Did it work in previous releases? At least according to this it sounds super experimental, so I don't think we'll block anything on it.

Alrlight, I will open an Issue there then. Probably just a small thing ...
And as side note: It's a new app for NC21 and was announced in the official release blog entry for NC21. It use binaries for x64, x86, arm64 and just "integrate" them to NC. The app does install everything automatically (also download and start the binaries using the NC app folder). So, it should work, because the user hasn't to setup anything manually (unlike the notify_push app).

szaimen:
> It use binaries for x64, x86, arm64 and just "integrate" them to NC.

It only ships binaries for 64bit x86 architecture, afaik. See https://help.nextcloud.com/t/nc-21-whiteboard-integration-error/109462/2

kyrofa:
Alright I think I'm happy with this, then I owe @szaimen a review.

PR #1740, opened 2021-05-23
Set compression to LZO

timlueg:
LZO compression could improve startup time. See https://ubuntu.com/blog/snap-speed-improvements-with-new-compression-algorithm

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -7,6 +7,7 @@ description: |
 
 grade: stable
 confinement: strict
+compression: lzo
 base: core18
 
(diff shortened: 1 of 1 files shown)


kyrofa:
Thanks for this @timlueg, but:

> To improve startup times, we decided to test a different algorithm – LZO – which offers lesser compression, but needs less processing power to complete the action.

I don't really want to force guinea pig status on all my users. Once Canonical decides to make LZO default, then we'll get the benefits along with (hopefully) stability. Until then, while I appreciate the work here, I think it's jumping the gun.

PR #1794, opened 2021-07-18
Add option to enable pretty URLS

szaimen:
Fix https://github.com/nextcloud/nextcloud-snap/issues/412

---

It works this way: every push that is pushed to any branch that is not called master, develop or '[12][0-9]' (this regex covers 10-29) will trigger the new workflow that downloads the current branch, runs the script that I wrote (My script basically sets up a temporary Nextcloud instance, enables pretty urls on it and writes all needed changes from there into a new .htaccess file.), changes the new .htaccess file to be based on the in the branch used nextcloud archive, commits the changes (if any) and pushes them to the same branch. This way you will be able to track any possible changes in the htaccess file in new Nextcloud versions in the future before merging into any of the mentioned branches.

If the feature gets enabled by the user it just includes my new .htaccess file in the apache conf which will automatically enable pretty urls for the users Nexcloud instance.

Signed-off-by: szaimen <szaimen@e.mail.de>

### DIFF
--- a/.github/workflows/get-prettyurl-conf.yml
+++ b/.github/workflows/get-prettyurl-conf.yml
@@ -1,0 +1,34 @@
+name: get prettyurl conf
+
+on:
+  push:
+    branches:    
+    - '**'
+    - '!master'
+    - '![12][0-9]'
+    - '!develop'
+
+jobs:
+  get-prettyurl-conf:
+    name: Get Prettyurl conf
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v2
+      - name: Create local changes
+        run: |
+          chmod +x ./prettyurl.sh
+          bash -x ./prettyurl.sh
+      - name: Commit files
+        if: ${{ success() }}
+        continue-on-error: true
+        run: |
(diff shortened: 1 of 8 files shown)


szaimen:
@kyrofa this is now much improved since it sets up a temporary Nextcloud instance, runs the normal command to enable pretty urls and writes the needed changes to the htaccess file. It is now much less of an hack than before and is hopefully integratable.

szaimen:
@kyrofa do you have some thoughts on this new attempt?

kyrofa (changes requested):
You're right @szaimen, this is improved, but it still has the same root problem of #1700: I'm not willing to land something that automatically pushes code to a PR. I believe PRs should be owned by the author as much as possible. I would much rather see this type of thing happening during the build of the snap instead of the approach currently taken.

By the way, there was no need to re-create this PR. Just push new commits!

szaimen:
I fear i will not be able to work on this anymore. But feel free to work on this @kyrofa and use my work as base for fixing the mentioned issue.

kyrofa:
Thanks @szaimen! I'll go ahead and close this for now, then.

PR #1880, opened 2021-10-05
Release 22.2.0snap1

kyrofa:
- apache: update to 2.4.50
- apache: support http compression
- nextcloud: update to 22.2.0
- php: update to 8.0.11
- mysql: disable performance_schema


README.md
```diff
@@ -4,7 +4,7 @@ Nextcloud server packaged as a snap. It consists of:
 
 - Nextcloud 22
 - Apache 2.4
-- PHP 7.4
+- PHP 8
```
r4co0n, line 7: Maybe PHP `8.0` would be better here, I think people are interested in the particular point release used, and 8.1 already has a RC.

r4co0n:
I just took a look over all of this, looks good except for the remarks about docs/comments, which are opinionated and are only brought up for consideration. (Excuse the 'Nevermind'-entry, I apparently am not allowed to resolve the review I brought up :/ )

r4co0n:
Please note that `text/html` is not part of the brotly-compressed mime-types, as currently configured, those files are still served as gzip-compressed. I did not follow the PR close enough to see if this is intentional, just want to bring it up.

kyrofa:
I'm noticing some issues with this on a fresh install. About 4 times out of 5 it just gets stuck spinning here:

Screenshot from 2021-10-06 09-28-26

I suspect this is a transient issue though, as it's happening on all releases back to v20. It must be trying to hit some online resource that is temporarily unavailable.

kyrofa:
> [...] an operator of the Nextcloud snap can expect to be rolling back to latest/stable, following the cancellation/close of short-lived sub-channel latest/beta/pr-1880, or am I mistaken?

No, it'll be the channel on which the branch is based, in this case `latest/beta`. To get back to stable you can revert or wait for release, when you can simply refresh.

Issue #1001, opened 2019-05-15
Unable to renew Letsencrypt certificate via SNAP ([Errno 1] Operation not permitted)


rjsears:
I am running snap 2.38.1 on Ubuntu 18.04 LTS. 

I installed nextcloud and set it up with letsencrypt and everything is running fine until it came time to renew my certificate which snap tried to do automatically (now multiple times). There seems to be some kind of OS or filesystem error and I am getting the following error in the log files each time it attempts to renew, except that it increases the privkey number (ex: privkey1.pem, privkey2.pem, etc) but the error is the same. THe LEtsEncrypt forum is saying that it is a problem with SNAP so I am looking for some help:

> 2019-05-15 16:19:18,454:DEBUG:acme.client:Storing nonce: LrlnuRVT0-aPA-PusulYmOvNuqdr7G8OxaAewYTF8do

> 2019-05-15 16:19:18,456:DEBUG:certbot.storage:Writing new private key to /var/snap/nextcloud/current/certs/certbot/config/archive/nextcloud.bob.aero/privkey4.pem.

> 2019-05-15 16:19:18,457:WARNING:certbot.renewal:Attempting to renew cert (nextcloud.bob.aero) from /var/snap/nextcloud/current/certs/certbot/config/renewal/nextcloud.bob.aero.conf produced an unexpected error: **[Errno 1] Operation not permitted: '/var/snap/nextcloud/current/certs/certbot/config/archive/nextcloud.bob.aero/privkey4.pem'. Skipping.**

> 2019-05-15 16:19:18,458:DEBUG:certbot.renewal:Traceback was:
Traceback (most recent call last):
  File "/snap/nextcloud/13144/lib/python2.7/site-packages/certbot/renewal.py", line 450, in handle_renewal_request
    main.renew_cert(lineage_config, plugins, renewal_candidate)
  File "/snap/nextcloud/13144/lib/python2.7/site-packages/certbot/main.py", line 1192, in renew_cert
    renewed_lineage = _get_and_save_cert(le_client, config, lineage=lineage)
  File "/snap/nextcloud/13144/lib/python2.7/site-packages/certbot/main.py", line 115, in _get_and_save_cert
    renewal.renew_cert(config, domains, le_client, lineage)
  File "/snap/nextcloud/13144/lib/python2.7/site-packages/certbot/renewal.py", line 315, in renew_cert
    lineage.save_successor(prior_version, new_cert, new_key.pem, new_chain, config)
[...]

PR #1901, opened 2021-10-13
WIP: Certbot: Bump version, use python 3, import webroot plugin from certbot

r4co0n:
When building the snap locally, I noticed that certbot is still built against Python 2. When inspecting all the py-files for compatibility with Python 3, I stumbled upon `src/https/certbot_nextcloud_plugin`. To avoid having to keep it up-to-date, I rewrote this script to simply import certbot's webroot plugin. I read the patch is no longer necessary, but kept for compatibility with old cli.ini files used in installations, see #1001.

Upon doing this, I ran into certbot erroring out because of some argparse initialization failure similar to certbot/certbot#7633. Then I found out about certbot installation via pip not being fully supported, and decided to try with the latest version of certbot, to see if I get a different error. Afterwards, I needed to fix/drop the relevant patches.

Now PR certbot/certbot#8131 introduced behaviour that results in:
> ERROR:certbot._internal.log:Exception: Duplicate plugin name webroot from certbot and nextcloud.

Just wanted to show you what I have until now to see if you are generally interested before I proceed.


r4co0n:
Dropping Python makes lots of sense, I will investigate acme.sh .

We will want to "recycle" the relevant snap configuration options so this just continues to work for people who have letsencrypt set-up, right?

PR #1945, opened 2021-11-21
Add Let's Encrypt command to README.md

noterminusgit:
The instructions to enable SSL using Let's Encrypt are currently missing from the README.md file, so I had to do some digging to find out how to enable it. This will add the appropriate command to the README.md file so others don't have to search for how to do it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ editing `/var/snap/nextcloud/current/nextcloud/config/config.php`), the snap
 exposes extra configuration options via the `snap set` command.
 
 
-#### HTTP/HTTPS port configuration
+#### HTTP/HTTPS configuration
 
 By default, the snap will listen on port 80. If you enable HTTPS, it will listen
 on both 80 and 443, and HTTP traffic will be redirected to HTTPS. But perhaps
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -81,6 +81,10 @@ Note that, assuming HTTPS is enabled, this will cause HTTP traffic to be
 redirected to port 444. You can specify both of these simultaneously as well:
 
     $ sudo snap set nextcloud ports.http=81 ports.https=444
+    
+To enable SSL (and automatic redirect from HTTP to HTTPS) using Let's Encrypt:
```
kyrofa, line 85: I'm not sure this anything beyond what is [already included in the README](https://github.com/nextcloud-snap/nextcloud-snap/blob/master/README.md?plain=1#L163-L167).

kyrofa (commented):
> The instructions to enable SSL using Let's Encrypt are currently missing from the README.md file

They aren't actually, but the fact that you looked and missed it means perhaps the README should be reorganized. Would it helped if we moved the "CLI utilities" section up near the top instead of where it currently is near the bottom?

README.md
gitfreebeer: I did indeed miss the commands you referenced. Thanks for the heads up. 

Yes, your suggestion to move the CLI utilities section up near the top would be helpful. Maybe also some text in the SSL section to reference the CLI utilities section for enabling SSL? I was a bit confused by the implication that I should know how to enable SSL or have it enabled already when I was reading the HTTP/HTTPS port configuration section. I believe it would be more complete if it had at least references to all the necessary HTTP/HTTPS configs.

gitfreebeer:
I did indeed miss the commands you referenced. Thanks for the heads up.

Yes, your suggestion to move the CLI utilities section up near the top would be helpful. Maybe also some text in the SSL section to reference the CLI utilities section for enabling SSL? I was a bit confused by the implication that I should know how to enable SSL or have it enabled already when I was reading the HTTP/HTTPS port configuration section. I believe it would be more complete if it had at least references to all the necessary HTTP/HTTPS configs.

Also, if the commands were formatted with the "  $ sudo" format, like the rest of the commands in the README, it would be more obvious how to use them.

Issue #1958, opened 2021-12-14
New major versions (atm 23.0.0) at beta or edge channel for testing


Pilzinsel64:
Hello!

Would it be possible to upgrade to the new major version 23.0.0 already now and keep it in an edge or beta channel until v23.0.1 has been released for testing? Would that even be possible in technical (thinking about because the automated build and release processes).

I would ask this question in generally for each upcomming major update. In my opinion it would help testing things and fix bugs even faster. Like me (who administrate and upgrade also other instances) would like to try to upgrade also the snap instance as they didn't regonice any big bugs or knowns the bugs and know how to deal with. The only option without this would be the latest/edge but don't wan't to directly use the master branche of nextclodu with possible NC24 changes.
So this would give people a change to test their setup and the new major version in the snap.

What do you think?

(I was thinking about doing a pull request for NC23, but not sure If "develop" is the right branche for since it's a new major version just for beta/edge channel. So, I just wait what you think about the idea.)

Pilzinsel64:
Closing, because see comment here: https://github.com/nextcloud-snap/nextcloud-snap/pull/1959#issuecomment-995130920

PR #1959, opened 2021-12-14
Update Nextcloud to 23.0.0

Pilzinsel64:
See https://github.com/nextcloud-snap/nextcloud-snap/issues/1958

Probably that's not the right way how-to but a first try. Don't know how to realize that it stays on edge or beta channel e.g.
Any suggestions are welcome 😅

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -188,8 +188,8 @@ parts:
 
   nextcloud:
     plugin: dump
-    source: https://download.nextcloud.com/server/releases/nextcloud-22.2.3.tar.bz2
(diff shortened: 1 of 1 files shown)


stondino00:
> See #1958
> 
> Probably that's not the right way how-to but a first try. Don't know how to realize that it stays on edge or beta channel e.g. Any suggestions are welcome 😅

Hey @Pilzinsel64!

Might be best to wait until 23.0.1 is released.  They already have a bunch of bugs fixed under that milestone.  No release date set though.

image

kyrofa:
Yeah generally we can get daily builds setup for 23/edge, but we don't want to land this on develop until it's at at least the .1 patch release.

PR #1966, opened 2021-12-27
apache config: use custom Diffie-Hellman parameters

### LINKED ISSUE #1660: Web server supports insufficiently secure parameters for Diffie-Hellman key exchange.
The following is copied from test results from Internet.nl:

> The security of elliptic curve Diffie-Hellman (ECDHE) ephemeral key exchange depends on the used elliptic curve. We check if the bit-length of the used elliptic curves is a least 224 bits. Currently we are not able to check the elliptic curve name.
> 
> The security of Diffie-Hellman Ephemeral (DHE) key exchange depends on the lengths of the public and secret keys used within the chosen finite field group. We test if your DHE public key material uses one of the predefined finite field groups that are specified in RFC 7919. Self-generated groups are 'Insufficient'.
> 
> The larger key sizes required for the use of DHE come with a performance penalty. Carefully evaluate and use ECDHE instead of DHE if you can.
> 
> Besides ECDHE and DHE, RSA can be used for key exchange. However, it is at risk of becoming insufficiently secure (current status 'phase out'). The RSA public parameters are tested in the subtest 'Public key of certificate'. Note that RSA is considered as 'good' for certificate verification.
> 
> See '[IT Security Guidelines for Transport Layer Security (TLS)](https://english.ncsc.nl/publications/publications/2019/juni/01/it-security-guidelines-for-transport-layer-security-tls)' from NCSC-NL, guideline B5-1 and table 9 for ECDHE, and guideline B6-1 and table 10 for DHE (in English).

Nextcloud's snap image should be using modern standards secp384r1, secp256r1, x448, and x25519 for better security. 

### **Proposed solution**
It comes down to generating a new DHE parameter and setting Apache to use it. 

I followed [this](https://raymii.org/s/tutorials/Strong_SSL_Security_On_Apache2.html#toc_4) guide to improve the `SSLOpenSSLConfCmd DHParameters` setting:

Generate the parameters
`openssl dhparam -out /etc/ssl/certsdhparam.pem 4096`

Add the following to your Apache config.
`SSLOpenSSLConfCmd DHParameters "/etc/ssl/certs/dhparam.pem"`

kyrofa:
This PR resolves #1660 by generating custom 4096-bit DH parameters. This needs to be tested on a variety of hardware to ensure there is no significant performance drop.


kyrofa:
Builds for this PR are available in `latest/beta/pr-1966`. Testing would be appreciated! @pachulo can you see how this behaves on a low-end device?

pachulo:
> @pachulo can you see how this behaves on a low-end device?
 
Hmmm, my production system runs on it @kyrofa , besides, I'm running `stable/21` on it, how could I test it without much hassle?

pachulo:
Gently ping @kyrofa ;)

kyrofa:
Thanks for the ping. I think there's no way around it being a huge hassle in your case, I'll get one setup here and take it for a spin, no worries.

Issue #542, opened 2018-05-02
Enable OCSP stapling


bn4t:
Enable OCSP stapling as it increases performance and privacy.
Is also recommended by https://mozilla.github.io/server-side-tls/ssl-config-generator/

Feel free to share any thoughts on this.

kyrofa:
I'm not so sure we want to deal with this. It sounds like the implementation in Apache [is a mess](https://blog.hboeck.de/archives/886-The-Problem-with-OCSP-Stapling-and-Must-Staple-and-why-Certificate-Revocation-is-still-broken.html).

bn4t:
Should I close issue, or leave it open?

Guillaume99:
> implementation in Apache is a mess.

quoted article has been updated, stating that "the situation has improved considerably"
maybe it is time now ?

pachulo:
OK, I've been checking this and to do it [the "better" way](https://github.com/icing/mod_md#stapling-in-mod_md) we need to use [this `mod_md` apache module](https://github.com/icing/mod_md):
* https://github.com/icing/mod_md#how-to-staple-all-my-certificates

I would like to try to implement it if you think it's worthwhile @kyrofa

pachulo:
gently ping @kyrofa

kyrofa:
It's interesting that mod_md might be considered the solution to this, @pachulo. It might tie together with the solution to #1902 as well, potentially. Stuff like this makes me nervous though:

> That being said, it is a new implementation. There will be bugs lurking and it is probably good advice to switch from the old stapling in a controlled way. Start with some domains and see how it works for you.

The idea of essentially making our users into unwitting beta testers makes me uncomfortable. When it breaks we're holding the pieces. What do you think?

pachulo:
> The idea of essentially making our users into unwitting beta testers makes me uncomfortable. When it breaks we're holding the pieces. What do you think?

Well, to be honest, I don't feel like doing it in that case. I don't really think it's such an important feature for us to rush.

pachulo:
[Let's Encrypt is droping OCSP support](https://letsencrypt.org/2024/12/05/ending-ocsp/), so we don't really need this anymore.

PR #1974, opened 2022-01-12
Enable OCSP stapling #542

Guillaume99:
proposed change to enable OSCP Stapling,

### DIFF
--- a/src/apache/conf/ssl.conf
+++ b/src/apache/conf/ssl.conf
@@ -42,6 +42,11 @@ Listen ${HTTPS_PORT}
 SSLProtocol all -SSLv3 -TLSv1 -TLSv1.1
 SSLProxyProtocol all -SSLv3 -TLSv1 -TLSv1.1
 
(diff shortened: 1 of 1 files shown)


pachulo:
Hi @Guillaume99 ! Thanks for this.

One question, did you try a compiled snap with this change somewhere?

Guillaume99:
> One question, did you try a compiled snap with this change somewhere?

unfortunately not, this is just a copy/paste from the mozilla SSL Configurator (https://ssl-config.mozilla.org/#server=apache&version=2.4.41&config=modern&openssl=1.1.1k&guideline=5.6)

Further more, i don't know what will happen for autogenerated certificate or non https user. Maybe it should be limited only to the letsencrypt users ?

pachulo:
Hi again!
I think that this can be closed and further discussion about how to implement this can go on on https://github.com/nextcloud-snap/nextcloud-snap/issues/542

Issue #1046, opened 2019-06-22
How to install ffmpeg?


szaimen:
I want to use the nextcloud video converter from the app store: https://apps.nextcloud.com/apps/video_converter
But it needs ffmpeg and is not installed in the snap. 
How can I do that?

kyrofa:
I'm afraid this isn't really possible as far as I can tell. I expect the snap itself would need to bundle ffmpeg, which isn't something we can do. If you can tell the app where to look for ffmpeg, perhaps you can build ffmpeg and install it into `/var/snap/nextcloud/current/` or some other place the snap can read (since it can't access the ffmpeg installed on your host due to confinement).

szaimen:
Because I do not know, if there is an option to configure the path where the app looks for ffmpeg, I created a new issue: https://github.com/PaulLereverend/NextcloudVideo_Converter/issues/18
When there was an option to configure the path, would a symbolic link to the normal installation path enough to get it working?

kyrofa:
I'm afraid not-- confinement follows symlinks, as it were. If the snap doesn't have access to the linked file, the symlink won't be useful either. It doesn't sound like there's a good solution here, I'm sorry.

kyrofa:
I'm going to go ahead and close this, but feel free to continue discussion here.

dmfgaming:
I have ffmpeg installed in snap, but yet it says its not installed. You can install ffmpeg in snap by sudo snap install ffmpeg

but The converter is still giving me, need to install ffmpeg, so how do I fix this?

kyrofa:
> sudo snap install ffmpeg

That just installs another snap containing ffmpeg; Nextcloud doesn't have access to it. My [comment above](https://github.com/nextcloud/nextcloud-snap/issues/1046#issuecomment-509676968) remains my only suggestion, I'm afraid.

Jachhj-sc:
> I'm afraid this isn't really possible as far as I can tell. I expect the snap itself would need to bundle ffmpeg, which isn't something we can do. If you can tell the app where to look for ffmpeg, perhaps you can build ffmpeg and install it into `/var/snap/nextcloud/current/` or some other place the snap can read (since it can't access the ffmpeg installed on your host due to confinement).

Thanks installing it in the current folder and updating source code for the app to take my path instead of asking the system where it is, worked.

shaman79:
[...]

PR #2013, opened 2022-02-26
Enable Movie files thumbnails creation

### LINKED ISSUE #1327: Video Thumbnails Failing to Generate on NextCloud Snap
I am running Ubuntu Server version 20.04 LTS and successfully installed NextCloud Snap; however, video thumbnails are not working. 

I installed FFmpeg and edited the config.php with the following text: 

```
'enable_previews' => true,
'enabledPreviewProviders' =>
array (
0 => 'OC\\Preview\\TXT',
1 => 'OC\\Preview\\Image',
2 => 'OC\\Preview\\Movie',
3 => 'OC\\Preview\\MP4',
),
```

But the same issue of not video thumbnails is occurring. Any assistance you might be able to provide would be incredibly helpful.

Thank you so much!

pachulo:
Adding `ffmpeg` and `libslang2` to `php` `stage-packages` seems to do the trick, at least on my limited testing.

With this, you only need to add something of the like to the `config.php`:
```
'enabledPreviewProviders' => [
    'OC\Preview\Movie',
],
```

[Because `Movie` is not in the default enabled list](https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/config_sample_php_parameters.html#previews):
> due to performance or privacy concerns

And video files thumbnails will be created on demand when using Files to access folders that contain them.

I've also trimmed some unnecessary, I think, folders from the installed packages. Maybe some other files can also be deleted.
The biggest file is `usr/lib/x86_64-linux-gnu/libLLVM-10.so.1` weighting 71M on `amd64`.

The size of the snap goes from 239M to 277M (almost a 16% increment) in that same arch.

Regarding the security record of ffmpeg: [is not ideal](https://ffmpeg.org/security.html) but I would say that it's better than ImageMagick's one, [at least regarding impact of vulnerabilities](https://www.cvedetails.com/vulnerability-list/vendor_id-3611/Ffmpeg.html). [Their attitude regards security vulnerabilities being disclosed also seems much better](https://security.googleblog.com/2014/01/ffmpeg-and-thousand-fixes.html).

Another thing far from ideal is that the packet [is not in `main`, but `universe`](https://packages.ubuntu.com/bionic-updates/ffmpeg).

But the main bright point for me is that this optional and only called if the admin enables the preview provider.

If accepted, this would fix #1327

Any input is very welcome! Thanks!

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -281,9 +281,17 @@ parts:
       - libzip4
       - libargon2-0
       - libonig4
+      # For video thumbnails
+      - ffmpeg
+      - libslang2
     prime:
      - -sbin/
      - -etc/
+     # ffmpeg related
+     - -usr/share/man/
+     - -usr/lib/x86_64-linux-gnu/dri/
+     - -usr/share/alsa/
+     - -usr/share/doc/
     organize:
       # Put php-fpm in bin with everything else, not sbin.
       sbin/php-fpm: bin/php-fpm


kyrofa:
> Regarding the security record of ffmpeg: [is not ideal](https://ffmpeg.org/security.html) [...].

I agree, I don't mind seeing CVEs. In fact, I would be _concerned_ if a project intended for consuming media files had no CVEs.

> [Their attitude regards security vulnerabilities being disclosed also seems much better](https://security.googleblog.com/2014/01/ffmpeg-and-thousand-fixes.html).

That's exactly what I would be looking for, which makes me feel far more comfortable with it. I do want to point out, though, that 

> Another thing far from ideal is that the packet [is not in main, but universe](https://packages.ubuntu.com/bionic-updates/ffmpeg).

That. That is the problem. For a project that handles files like this, there will be security issues. They seem to have an established path and history of handling those. But if the package is in universe, we won't actually get any of their fixes. Which leads us down the build-from-source path, which sounds awful :( .

pachulo:
> That. That is the problem. For a project that handles files like this, there will be security issues. They seem to have an established path and history of handling those. But if the package is in universe, we won't actually get any of their fixes.

I totally understand your concerns. There seems to be [some](https://ubuntu.com/security/notices/USN-4431-1) [fixes](https://ubuntu.com/security/notices/USN-3967-1) that find they way to the package, but it seems a best effort thing, to be honest.

> Which leads us down the build-from-source path, which sounds awful and will add to our build times and maintenance burden.

Then, I think we might have other options, and I would like your opinion on them. Mainly:
1. [Use this well established PPA](https://launchpad.net/~savoury1/+archive/ubuntu/ffmpeg4), which packs the latest 4.x stable version.
2. [Use the static builds that this folk creates](https://johnvansickle.com/ffmpeg/), [that seems to be quite official](https://ffmpeg.org/download.html).

I do prefer the latest over the former, as we should only download the [static build for the version we want](https://www.johnvansickle.com/ffmpeg/old-releases/), and keep only the executable we need.
Furthermore, there are builds for `i686`, `armhf`, `armel`, `arm64` and `amd64` archs, which should cover everything the snap can be compiled against, won't it?

pachulo:
Hi @kyrofa ! I want to start testing the second option mention above, but first I want to know what you think about it.Thanks!

pachulo:
gently ping @kyrofa

kyrofa:
@pachulo the static build is interesting, but it makes my mind go a different direction with this. Have you looked into how the preview providers find and use ffmpeg? If there are established static builds we could conceivably write a wiki page about how to download it into `$SNAP_DATA` somewhere and point the preview provider at it-- then folks who want it can get previews for these files, without everyone having to take on the cost of shipping (and in our case maintaining) it.

pachulo:
> Have you looked into how the preview providers find and use ffmpeg?

Not very deeply, but I will.

> without everyone having to take on the cost of shipping (and in our case maintaining) it.

In the case that the option you commented above is not viable, I am comfortable taking the responsibility of maintaining it, and I also think that the cost of shipping a fatter snap vs not having this functionality is worth it.

pachulo:
gently ping @kyrofa ;)

DillFrescott:
So glad this might finally be added! Yay!

lovekyleman:
Please add it. Really need it

kyrofa:
Sorry for the delay @pachulo.

> OK, it seems that nextcloud only looks for ffmpeg in the $PATH of the Apache server, and it's location cannot be configured, so I guess that we cannot do what you say here, can't we?

Any chance you could point me to where you found that? I'm really not comfortable pulling unvetted binaries into this thing, I'd like to continue exploring allowing folks to opt in with those static builds, even if it means proposing a feature upstream.

pachulo:
Hi @kyrofa ,

> Any chance you could point me to where you found that?

I don't remember exactly, but I think that it was in here https://github.com/nextcloud/server/blob/stable23/lib/private/Preview/Movie.php
And I also searched in the documentation, but bear in mind that my PHP-foo is quite weak and I might be missing something...

> I'm really not comfortable pulling unvetted binaries into this thing, I'd like to continue exploring allowing folks to opt in with those static builds, even if it means proposing a feature upstream.

That would be absolute great!

pachulo:
I was thinking more about this....@kyrofa do you think it would be an option to create a snap that only packaged the `ffmpeg` static binary and then exposed a `slot` for the nextcloud snap to connect?


I took the idea from here:
* https://forum.snapcraft.io/t/using-chromium-ffmpeg-in-third-party-browser-snaps/6545/1
* https://git.launchpad.net/~chromium-team/chromium-browser/+git/chromium-ffmpeg/tree/snapcraft.yaml
* https://snapcraft.io/docs/content-interface

pachulo:
Gently ping @kyrofa !

nonetrix:
What is the best way to apply this patch to my server? I really don't want to have to reinstall NextClould I assume I would have to use this before it's merged if at all though

pachulo:
Hey @kyrofa ! Any thoughts on the proposed solution in https://github.com/nextcloud-snap/nextcloud-snap/pull/2013#issuecomment-1146439368

If it's a yes I would like to try that approach whenever I have time. And I think that I could also pack and maintain the static builds ffmpeg snap.

Thanks!

ngvoicu (approved):

ngvoicu:
Please merge it.

bobboau:
A snap got pushed recently and looks like something changed with the thumbnails,  they now have an animated thumbnail that looks like a "loading" graphic, but they still are not working properly, I'm hoping we can get this soon?

Deniffer:
I'd like to keep track of this thread update, since it is truely important for us to have a video thumbnails  on a snap version.  i believe that for most of us storing our data using nextcloud would have this demand. sincerely appreciate the work that @pachulo does, it would be helpful for reviewing this PR soon . Gently remind @kyrofa

shamiv:
> I'd like to keep track of this thread update, since it is truely important for us to have a video thumbnails on a snap version. i believe that for most of us storing our data using nextcloud would have this demand. sincerely appreciate the work that @pachulo does, it would be helpful for reviewing this PR soon . Gently remind @kyrofa

I have been closely following this thread as well, and also hope this will be fixed soon!

traeblain:
@kyrofa Based on your release schedule, it looks like there's one coming up soon.  Hoping this makes it into the next release.  This is the primary item keeping me from mainlining the snap version.

szaimen:
This might not be needed after all?
See https://github.com/nextcloud-snap/nextcloud-snap/issues/1046#issuecomment-1374427458

kyrofa:
> I was thinking more about this....@kyrofa do you think it would be an option to create a snap that only packaged the ffmpeg static binary and then exposed a slot for the nextcloud snap to connect using a plug if the user executed a command?

@pachulo, that idea has potential. If we do it right, we could develop a generic solution to "I need the snap to be able to access X binary". We've had similar requests for, say, clamAV. There are various complications with the content sharing approach in terms of the LD_LIBRARY_PATH, but for static binaries it should work pretty well.

kyrofa:
@szaimen sadly not all apps support specifying where their ffmpeg is. They expect it to be on the PATH. There is still work to do, here.

szaimen:
> @szaimen sadly not all apps support specifying where their ffmpeg is. They expect it to be on the PATH.

which apps exaclty? 

So maybe creating a PR in server that allows to specify a path to the ffmpeg binary instead of sniffing it would make sense?

kyrofa:
I'm afraid I don't know, I'm just parroting @pachulo who has put all the effort into this. Although a quick glance at the movie app leads me to believe it [actually might be configurable](https://github.com/nextcloud/server/blob/stable25/lib/private/Preview/Movie.php#L68). That's the one that generates thumbnails, right?

szaimen:
> That's the one that generates thumbnails, right?

Yes

ZOKOB:
Hello I followed your tutorial however I still don't have my video thumbnails here is my configuration file

  'memories.exiftool' => '/var/snap/nextcloud/33730/nextcloud/extra-apps/memories/exiftool-bin/exiftool-am
  'memories.vod.ffmpeg' => '/var/snap/nextcloud/bin/ffmpeg',
  'memories.vod.ffprobe' => '/var/snap/nextcloud/bin/ffprobe',
  'memories.vod.disable' => false,
  'memories.ffmpeg_path' => '/snap/ffmpeg-integration-gplv3/1/bin/ffmpeg',
  'memories.ffprobe_path' => '/snap/ffmpeg-integration-gplv3/1/bin/ffprobe',
  'memories.tmp_path' => '/var/tmp',
  'memories.vod.path' => '/var/snap/nextcloud/33908/nextcloud/extra-apps/memories/exiftool-bin/go-vod-amd64',
  'preview_ffmpeg_path' => '/var/snap/nextcloud/bin/ffmpeg',
  'enable_previews' => true,
  'enabledPreviewProviders' =>
   array (
    0 => 'OC\Preview\TXT',
    1 => 'OC\Preview\MarkDown',
    2 => 'OC\Preview\OpenDocument',
    3 => 'OC\Preview\PDF',
    4 => 'OC\Preview\MSOffice2003',
    5 => 'OC\Preview\MSOfficeDoc',
    6 => 'OC\Preview\Image',
    7 => 'OC\Preview\Photoshop',
    8 => 'OC\Preview\TIFF',
    9 => 'OC\Preview\SVG',                                                            10 => 'OC\Preview8\Font',
   11 => 'OC\Preview\MP3',
   12 => 'OC\Preview\Movie',
   13 => 'OC\Preview\MKV',
   14 => 'OC\Preview\MP4',
   15 => 'OC\Preview\AVI',
   )
);`
can you help me please?

jkoehler11:
I haven't been able to get this to work either. I've been trying to get both Video Previews and the Memories App to work.

 I've added the static binaries to `/var/snap/nextcloud/current/bin/`

```bash
@ubuntu:/var/snap/nextcloud/current/bin$ ls -lrt
total 98104
-rwxr-xr-x 1 root root 50015760 Apr  7 22:02 ffmpeg
-rwxr-xr-x 1 root root 49946640 Apr  7 22:02 ffprobe
```

Then in the config:
```
  'memories.vod.ffmpeg' => '/var/snap/nextcloud/current/bin/ffmpeg',
  'memories.vod.ffprobe' => '/var/snap/nextcloud/current/bin/ffprobe',
  'memories.vod.disable' => false,
  'memories.ffmpeg_path' => '/var/snap/nextcloud/current/bin/ffmpeg',
  'memories.ffprobe_path' => '/var/snap/nextcloud/current/bin/ffprobe',
  'memories.tmp_path' => '/var/tmp',
  'preview_ffmpeg_path' => '/var/snap/nextcloud/current/bin/ffmpeg',
  'enable_previews' => true,
  'enabledPreviewProviders' => 
  array (
    0 => 'OC\\Preview\\PNG',
    1 => 'OC\\Preview\\JPEG',
    2 => 'OC\\Preview\\GIF',
    3 => 'OC\\Preview\\BMP',
    4 => 'OC\\Preview\\XBitmap',
    5 => 'OC\\Preview\\Movie',
    6 => 'OC\\Preview\\MP3',
    7 => 'OC\\Preview\\TXT',
    8 => 'OC\\Preview\\MarkDown',
    9 => 'OC\\Preview\\MP4',
  ),
```

When I create the memories index, it says that mpeg and mp4 are not supported.

```bash
@ubuntu:/var/snap/nextcloud/current/bin$ sudo nextcloud.occ memories:index
Checking database indices

MIME Type support:
  image/png: supported
  image/jpeg: supported
  image/heic: not supported
  image/png: supported
  image/tiff: not supported
  image/gif: supported
  image/bmp: supported
  image/x-dcraw: not supported
  video/mpeg: not supported
  video/webm: not supported
  video/mp4: not supported
  video/quicktime: not supported
  video/x-matroska: not supported
  video/MP2T: not supported

Some file types are not supported by your preview provider.
```

bobboau:
Is there any motion on this? It's been years now and because this is in a snap I have no ability to fix this myself. Is there some sort of workaround? Is this static linking of ffmpeg a viable solution? Is there some alternative solution coming? If so what is the status on that? I really don't want to see this be one of those things that stays broken for half a decade. Especially when I see someone did the open source thing and provided a fix themselves. Every time I use nextcloud I see this bug and it makes me wonder if I made a mistake jumping to use the snap version. Wondering what would be involved in migrating my files to a different instalation method that I could fix problems like this myself with. Is there any intention to fix this? Or is this not a priority? Should I give up?

dagguh:
@pachulo gotta rebase, right?

ZOKOB:
Good evening,

I would like to know if this functionality will be provided in a next nextcloud update?

Or do I have to switch to standard installation?

Thanking you in advance, I wish you a pleasant evening.

dagguh:
Let's ship it 🙂

lovekyleman:
Come on baby do it

szaimen:
> This should not be needed with nc26 anymore due to [nextcloud/server#36225](https://github.com/nextcloud/server/pull/36225). However it will need some documentation in the wiki.

bobboau:
> > This should not be needed with nc26 anymore due to [nextcloud/server#36225](https://github.com/nextcloud/server/pull/36225). However it will need some documentation in the wiki.

Has this documentation been added?
If so does anyone have a link to it?

PR #2088, opened 2022-05-23
Update phpredis 5.3.7

pachulo:
I guess we should be updating this too, now that https://github.com/nextcloud-snap/nextcloud-snap/pull/2084 was merged.


kyrofa (approved):

kyrofa:
Yeah they don't really coordinate releases as far as I'm aware, but definitely worth having them both up to date. Thank you!

Issue #2024, opened 2022-03-16
Update Netxcloud 23.0.2snap1 -- error message PHP OPcache module is not properly configured:


scubamuc:
After snap update to Netxcloud 23.0.2snap1

an error message appears in settings/overview

> 
> The PHP OPcache module is not properly configured:
> 
> The OPcache interned strings buffer is nearly full. To assure that repeating strings can be effectively cached, it is recommended to apply opcache.interned_strings_buffer to your PHP configuration with a value higher than 8.

How can this be corrected in cofig.php?
Which value is suggested?

kyrofa:
Don't worry, this isn't harmful. Keep an eye on it over the next few days-- do you continue seeing it? If so we can make the interned buffer cache size configurable so you can get a little more performance out of it.

scubamuc:
Restarted nextcloud-snap `# snap restart nextcloud ` to see message disappear!
Please close and forget

kyrofa:
Right, this is a cache in PHP-- restarting clears it. As you use Nextcloud, it fills up and swaps strings out. If it gets pretty full, the new release of Nextcloud warns you about it. If you see a warning constantly, it means you would benefit from slightly improved performance if you increased the cache size (something that is not configurable in the snap today). So please still keep an eye on it-- if you see it often, I'm willing to make that configurable.

scubamuc:
Thanks, I'll keep an eye on it nonetheless and give you feedback.
Configurable is great!

scubamuc:
warning message continues.
I have several (4) systems running the same snap with almost identical setups. 
Only one system displays the warning message; 
![grafik](https://user-images.githubusercontent.com/54933878/158952841-6618df78-5614-431a-864b-0c2460d57210.png)
Please make configurable.

ourownstory:
Same issue here. 
Please allow us to set
``` 
opcache.memory_consumption
opcache.interned_strings_buffer
```
Thank you!

kyrofa:
Alright, I'm on board with this, just need to find time to do it unless someone else wants to take a crack at it.

aagostini:
Hi to all,
I see the same warning message :
_The OPcache interned strings buffer is nearly full. To assure that repeating strings can be effectively cached, it is recommended to apply opcache.interned_strings_buffer to your PHP configuration with a value higher than 8._

on my two snap instance: **[Nextcloud Hub II](https://nextcloud.com/) (23.0.3)**

[...]

PR #2216, opened 2022-10-27
opcache settings

kz6fittycent:
Testing for the following:

See #2024 

Set opcache interned strings buffer
Set opcache memory consumption

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,19 +196,31 @@ There are a few CLI utilities included:
     - Nextcloud data
 
 #### What about ZFS?
+
 - Zpools may be bind mounted to accomplish data access by the snap 
     - For example, create a new directory (e.g. `mkdir -p /mnt/zpool_nc`)
     - Depending on where your zpool exists (e.g. `/pool/tank01`), you can create a
       bind mount to allow the snap to acces/manage your data
-        - For example: `mount --bind /mnt/zpool_nc /pool/tank01`
+        - For example: `mount bind -o /mnt/zpool_nc /pool/tank01`
     - NOTE: This requires that `removable-media` be connected:
         - `sudo snap connect nextcloud:removable-media`
     - You'll need to edit your `config.php` file to ensure that your data 
       directory is present in the configuration:
         - Edit `/var/snap/nextcloud/current/nextcloud/config/config.php` to note the
           data location (e.g. `'datadirectory' => '/mnt/zpool_nc',`)
-        - Ensure that the `.ocdata` file exists within the directory
+        - Ensure that the `.ocdata` file exists within the directory (if warnings are seen)
           (e.g. `sudo touch /mnt/zpool_nc/.ocdata`)
+        - Ensure that your fstab entries are correct:
+          For example:
+             ```
+             /pool/tank01 /mnt/zpool_nc	none	bind,x-systemd.requires=zfs-mount.service	0 0 
+             ```
+             
+             For more info on bind mounting ZFS zpools, please reference:
+             https://wiki.archlinux.org/title/ZFS#Bind_mount (or other reputable source).
+             
+             Also see `fstab` man pages for more info: `man fstab`
+             
     - Restart the snap `sudo snap restart nextcloud`
 
 ## Hacking
--- a/README.md
+++ b/README.md
@@ -195,6 +195,21 @@ There are a few CLI utilities included:
 - `$SNAP_COMMON` (`/var/snap/nextcloud/common/` by default)
     - Nextcloud data
 
+#### What about ZFS?
+- Zpools may be bind mounted to accomplish data access by the snap 
+    - For example, create a new directory (e.g. `mkdir -p /mnt/zpool_nc`)
+    - Depending on where your zpool exists (e.g. `/pool/tank01`), you can create a
+      bind mount to allow the snap to acces/manage your data
+        - For example: `mount --bind /mnt/zpool_nc /pool/tank01`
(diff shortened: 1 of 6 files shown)


src/php/config/php.ini
```diff
@@ -1769,11 +1769,16 @@ opcache.enable=1
...
+# opcache.interned_strings_buffer
+# See https://github.com/nextcloud-snap/nextcloud-snap/issues/2024
+
 ; The OPcache shared memory storage size.
-;opcache.memory_consumption=128
+opcache.memory_consumption=128
 
 ; The amount of memory for interned strings in Mbytes.
-;opcache.interned_strings_buffer=8
+opcache.interned_strings_buffer=8
```
kyrofa, line 1781: The rubber doesn't seem to be meeting the road here: no matter what the user sets, these values will always be 128/8 until you start using the values that they set. We use [environment variables](https://github.com/nextcloud-snap/nextcloud-snap/blob/a5868d9bc03315cab14bdab6e29e5efeaf8c78e0/src/php/config/php.ini#L430) for that [elsewhere](https://github.com/nextcloud-snap/nextcloud-snap/blob/a5868d9bc03315cab14bdab6e29e5efeaf8c78e0/src/php/bin/run-php#L8).

src/php/config/php.ini
```diff
@@ -1769,11 +1769,16 @@ opcache.enable=1
 ; Determines if Zend OPCache is enabled for the CLI version of PHP
 ;opcache.enable_cli=0
 
+# Set default values for: 
+# opcache.memory_consumption
+# opcache.interned_strings_buffer
+# See https://github.com/nextcloud-snap/nextcloud-snap/issues/2024
```
kyrofa, line 1775: No need to refer back to the issue. The commit/git history will do that for us.

src/hooks/bin/configure
```diff
@@ -92,6 +92,61 @@ handle_php_memory_limit()
 	snapctl restart nextcloud.php-fpm
 }
 
+# Set the following values - kz6fittycent
+# opcache.memory_consumption
+# opcache.interned_strings_buffer
+# See https://github.com/nextcloud-snap/nextcloud-snap/issues/2024
```
kyrofa, line 98: Likewise, this comment block is unnecessary.

README.md
```diff
@@ -195,6 +195,33 @@ There are a few CLI utilities included:
 - `$SNAP_COMMON` (`/var/snap/nextcloud/common/` by default)
     - Nextcloud data
 
+#### What about ZFS?
```
kyrofa, line 198: Please remove this here, looks like leftovers from another PR :smile: .

src/php/config/php.ini
kz6fittycent: ok - do I just need to comment those out again? Will user input uncomment and set the new values?

src/php/config/php.ini
kz6fittycent: ok - do I just need to comment those out again? Will user input uncomment and set the new values?

kyrofa: Take a look at the links I gave there-- you'll need to define environment variables holding the values set by the user, and then use those environment variables here in the config. Don't forget the php-fpm config as well.

ShapeShifter499:
Any updates with this?

kz6fittycent:
None yet. Life is busy right now and I'm heading out to visit family. I'll get another chance to revisit this once Thanksgiving is over.

YoannPa:
Maybe I should have posted my [comment](https://github.com/nextcloud-snap/nextcloud-snap/issues/2024#issuecomment-1321220785) here.
@kz6fittycent could you have a look at the list of parameters I share in order to make them available too ?
Thank you.

src/php/config/php.ini
szaimen, line 1781: Btw, just wanted to mention that when opcache.memory_consumption is set to the default of 128, setting opcache.interned_strings_buffer to 32 would always hide the warning. See https://github.com/nextcloud/server/pull/32902/files

szaimen: And based on my own observation is ir nearly impossible to make the opcache memory of 128m full with a nextcloud instance

szaimen: So what I wanto to say is when opcache.memory_consumption is set to 128 and opcache.interned_strings_buffer is set to 32, I am not sure if a config value is needed

kyrofa: Totally missed this until now-- that's brilliant @szaimen, thank you!

kyrofa: Proposed in #2301.

szaimen: LGTM 👍

kyrofa:
Closing in favor of #2301.

src/php/config/php.ini
THE-cmcurtin: > Btw, just wanted to mention that when opcache.memory_consumption is set to the default of 128, setting opcache.interned_strings_buffer to 32 would always hide the warning. See https://github.com/nextcloud/server/pull/32902/files

Two years later this arises. It seems like even if we don't have the option to configure every detail, we should be able to tweak one of the variables such that it will adjust the others accordingly.

I've got an instance now showing:
The PHP OPcache module is not properly configured. The OPcache buffer is nearly full. To assure that all scripts can be hold in cache, it is recommended to apply "opcache.memory_consumption" to your PHP configuration with a value higher than "128".. For more details see the documentation ↗.

PR #2222, opened 2022-11-03
Upgrade Nextcloud to 25.0.1

### LINKED ISSUE #2211: Upgrade to Nextcloud 25
Nextcloud released v25 a few days ago. The snap is a bit behind currently.

stondino00:
Fixes #2211

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -188,8 +188,8 @@ parts:
 
   nextcloud:
     plugin: dump
-    source: https://download.nextcloud.com/server/releases/nextcloud-24.0.6.tar.bz2
-    source-checksum: sha256/b26dff9980a47e7e722805fdbbf87e07f59a3817b03ecc32698e028e9baf0301
+    source: https://download.nextcloud.com/server/releases/nextcloud-25.0.1.tar.bz2
+    source-checksum: sha256/72d4076924caf19139c40178597af6211799e20440ce196fb43b9c4e47d77515
     organize:
       '*': htdocs/
       '.htaccess': htdocs/.htaccess


Pilzinsel64:
Not sure what's up with the test, but I had a quick look to it just for personal interest.
Seems that it is not able to find the user field anymore?
```
Failure/Error: fill_in "User", with: "anything-you-want"
```

Nextcloud 25:
grafik
grafik

Nextcloud 24:
grafik
grafik

szaimen:
I would not merge this as there is a bug that breaks 25 for 32-bit.
This will be fixed for 25.0.2.

kyrofa:
Thanks @szaimen, any chance you know what issue that is so I can monitor it?

szaimen:
It is alredy fixed for 25.0.2. See https://github.com/nextcloud/server/pull/34905

kyrofa:
Ah excellent, thank you. Okay, we will hold off on this indeed.

Pilzinsel64:
Oh sorry, just saw the "latest-25.tar.bz2", forgot about the "latest-stable..." files. Well, maybe soon...

Pilzinsel64:
By the way, just for personal interest, why using the tarball from the daily directory instead of the releases directory?
Without knowing the reason: Not sure if it is a good idea to relay on a `.../daily/latest-stableXY.tar.bz2` instead of `.../releases/latest-XY.tar.bz2`? Because of that it probably will never be possible to do early testing, I think.

kyrofa:
> By the way, just for personal interest, why using the tarball from the daily directory instead of the releases directory?

Because all we're really plumbed for are dailies. We generate a new snap every day from those tarballs for every release we support so we (and others) can test them. Whereas creating a new snap from a release we don't support yet is a manual affair.

szaimen:
Why not use the `latest-25` file then?

kyrofa:
Because it doesn't change on a schedule. We would never know when to spin a release. We try to avoid manual work as much as possible.

kyrofa:
Well, no. latest-stableXX is the latest commits slated for release. latest-XX is the latest release. A daily build of the former is different every day. A daily build of the latter is a bunch of identical snaps wasting bandwidth and storage space.

szaimen:
Anyway, was just an idea

szaimen:
The 25.0.2 update is finally out!
https://nextcloud.com/blog/maintenance-releases-25-0-2-24-0-8-and-23-0-12-are-out/

kyrofa:
Closing in favor of #2247.

PR #2229, opened 2022-11-08
Update opcache.interned_strings_buffer

bonnedav:
The security and setup warnings page in the current version throws a warning about opcache configuration and recommends that opcache.interned_strings_buffer be defined with a value higher then 8 to assure that repeating strings can be effectively cached.


kyrofa:
Whether one sees this message or not is entirely dependent on how heavily one's instance is used. There is no single catchall setting we can apply here-- it needs to be configurable. Closing in favor of #2216.

PR #2253, opened 2022-12-15
added support for export-directory for nextcloud.export

knorrfix:
Hallo,

I'm currently moving my nextcloud instance to a new server. I noticed it is currently not possible to export the export-data to a custom folder. 

I wrote this pull request in the meantime while waiting for the copy of my data to an drive. So I can carry it to the new Server. I think this Option will have a lot of usage and timesaving for the users. In case of moving the data to another server you always need to copy the data.

I'm relatively new to shell-script so I hope my modifications to the export-script satisfy your standards.

### DIFF
--- a/src/import-export/bin/export-data
+++ b/src/import-export/bin/export-data
@@ -22,7 +22,7 @@ FORMAT="1"
 print_usage()
 {
 	echo "Usage:"
-	echo "    $COMMAND [OPTIONS]"
+	echo "    $COMMAND [OPTIONS] [backup dir]"
 	echo "    Export data suitable for migrating servers. By default this"
 	echo "    includes the Nextcloud database, configuration, and data"
 	echo "    (equivalent to running $COMMAND -abcd)."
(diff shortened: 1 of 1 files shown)


kyrofa:
We've had discussions about this in the past-- it would definitely be useful, but the reason the functionality doesn't already exist is because the snap actually doesn't have permission to write to _most_ places (by default it can only really write to one place: where it's already writing the backup). You can't just hand it a directory and have it write to it, because it's confined. Which leads to a terrible user experience: "Why can't you write there?! It's owned by the right people... why is permission denied?!" Most users don't understand confinement, so this kind of option would just be a footgun.

What we need instead is a way to stream a tar to stdout, where the shell can be used to redirect it wherever the user wants, but it has proven to be a difficult problem to solve. Is that something you'd like to investigate?

scubamuc:
I don't know if this is redundant, but one of the selling points of using snap is the possibility to move the snaps between different machines without complication:

Your suggestion seems to be the long way round... which is also manageable and even scriptable. 
https://github.com/scubamuc/scubamuc.github.io#3-export-import----backup-restore

Below is the "snap" way to migrate and describes a snap container export. that should be the easiest solution, as the entire snap would be transferred to the new machine including DB, Data, and Certs!
https://github.com/scubamuc/scubamuc.github.io#32-nextcloud-snap-snapshot--restore

This even works when migrating an entire LXD instance including all containers and settings to another machine. IMO snap is the way to go for easy maintenance and simple migration.

I'm with @kyrofa here, Snap in general and Nextcloud-snap have great export, backup and migration solutions. Why reinvent the wheel?

PR #2254, opened 2022-12-17
integration tests: use input ID instead of label

kyrofa:
The login labels changed in Nextcloud v25 such that these tests no longer pass. Start using the IDs instead, which remain consistent through at least the last few releases.


Pilzinsel64:
Good work! 🙂👍
Just wondering, I mean, I tried to change the upper case "User" to lowercase "user" and password but still failed. That system is very new to me. 😅

kyrofa:
Well, we'll see if CI likes it. These passed locally. Taking a look at the commit you tried, it looked like you only updated the login tests, but that logic is used in a few others that would still fail.

PR #2247, opened 2022-12-09
nextcloud: update to 25.0.2

Pilzinsel64:
Fix #2211 
Fix #2222 

Don't want to attacked you @stondino00, just had the time for yet. Feel free to close this PR if not needed.

Anyways, not sure, if the tests success, as mentoined in [this comment](https://github.com/nextcloud-snap/nextcloud-snap/pull/2222#issuecomment-1302154488). No clue how fix them myself right now.

Also, not sure if it's right to change the changelog already here. Not sure what your process is for. Just a suggestion from my side.

PR #2256, opened 2022-12-18
integration tests: wait for nextcloud fixer to finish

kyrofa:
We have some tests that reinstall the snap, in which case Nextcloud Fixer runs, which is something that causes maintenance mode to be enabled. If the timing is Just Right (or Wrong, as the case may be), maintenance mode will be enabled right as a test is trying to log in, which causes the test to fail. Just wait for the Fixer to finish before trying anything.

### DIFF
--- a/tests/integration/import_export_spec.rb
+++ b/tests/integration/import_export_spec.rb
@@ -28,6 +28,7 @@ feature "Import and export data" do
 		`sudo mv "#{moved_backup}" "#{backup}"`
 		`sudo nextcloud.import "#{backup}"`
 		wait_for_nextcloud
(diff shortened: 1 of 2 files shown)


kyrofa:
@Pilzinsel64 I believe I've gotten to the bottom of the test failure in #2247 over here, if you want to take a look.

Pilzinsel64:
Would absolutely make sense. In NC25 the fixer needs longer for a very small amount of time due one or two additional steps in of the Nextcloud Updater (not sure what, but there is at least one new one) and then the test failes.

kyrofa:
Ah, that explains it!

PR #2303, opened 2023-02-07
Upgrade Boost

### LINKED ISSUE #2302: Upgrade Boost
Contains bugfixes and appears to help against some vulnerabilities that may not be directly in the boost code.  Just keeping things updated.

stondino00:
Fixes #2302


kyrofa:
I believe mysql actually has a pretty hard versioning requirement for boost. Indeed, this appears to not build:

Screenshot from 2023-02-07 08-34-20

PR #2364, opened 2023-03-29
Better document nextcloud.import in README

dsofeir:
There is essential advice about restoring backups contained in the [wiki](https://github.com/nextcloud-snap/nextcloud-snap/wiki/How-to-backup-your-instance). Without this information I spent 40 minutes trying to understand vague rsync errors. I think other users would benefit by this information being placed in a more prominent position.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,10 @@ There are a few CLI utilities included:
     - Import data exported from another Nextcloud snap instance (via
       `nextcloud.export`). By default this imports the database, config, and
       data. See `nextcloud.import -h` for more information. Note that it
-      requires `sudo`.
+      requires `sudo`. The path needs to be available from confinement, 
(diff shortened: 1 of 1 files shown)


kyrofa (changes requested):
Good call. However, the README is designed to be more of a brief overview rather than all-encompassing. How would you feel about putting it in the [help text of the command itself](https://github.com/nextcloud-snap/nextcloud-snap/blob/0c472ebaaa9ddda00c52805ae5c0e2c12042dfdb/src/import-export/bin/import-data#L20)?

dsofeir:
I think the current problem is that discovery is hard. I checked the README and command help text and saw no mention it was a reddit post that guided me to the Wiki.

Given this, yes, I am in favour of putting it either in the README or command help text. Either would be helpful.

kyrofa:
Yeah I'd prefer to see it in the help text, please.

Pilzinsel64:
Closing as already resolved in #2433

PR #2431, opened 2023-06-14
php: add config flag --enable-sysvsem

### LINKED ISSUE #2365: [NC26+] Add PHP-Module `sysvsem`
Nextcloud 26 suggest `sysvsem` as recommended php module for performance reason.
By reading [this](https://help.nextcloud.com/t/26-0-php-extension-sysvsem/158819/5) it is used to limit the amount of previews generated at the same time.

I suggest to add this php module.

grafik

_(Seen that after updated our prod to `26/edge` for early testing - except this warning everthing seems to be correct and working! Just as an early feedback.)_

Pilzinsel64:
Fix #2365

If you don't like the new module, just close this PR. ;)

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -228,6 +228,7 @@ parts:
       - --disable-rpath
       - --enable-ftp
       - --enable-bcmath
+      - --enable-sysvsem
 
       # Enable ldap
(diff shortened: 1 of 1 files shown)


kyrofa (approved):
Thank you @Pilzinsel64, I'm on board with adding core Nextcloud dependencies indeed.

PR #2468, opened 2023-07-21
nextcloud: enable theming by default

### LINKED ISSUE #2456: Enable Theming by default
As I regoniced during #2452: What do you think should we do with the thiming app? In Nextcloud since 26 (or even 25?) it is a hard dependency and you can't even disable the app on the normal way with the message "This app can not be disabled.". It isn't even listed on the Web Interface as installable app anymore, like the files app e.g.

Should we go ahead and also enable theming for snap by default? I mean, it is a core feature and were an big feature for the NC25 update and is also used for accessibility, if I'm right. Not sure if keep it disabled is the best anymore. I mean, if someone has problems due theming we can navigate him to the the upstream repo, same as with other core issues. What do you think? 🤔 

I opened an issue for to discuss that. Maybe we can share our experiences with. On my side it mostly works fine.

_Originally posted by @Pilzinsel64 in https://github.com/nextcloud-snap/nextcloud-snap/issues/2452#issuecomment-1632975067_

Pilzinsel64:
Fix #2456 (Only for new instanced)

Not sure if I should remove the script completely or keep as leftover. (It is named as "1_..." 😅)
I guess this is the only place here. Otherwise just tell me. :)

Should we also force enable on existing instances? If yes, I would add another script for.
I think it's ok to not enable on upgrade. The user maybe don't want it and knows what he do.

EDIT: Closed for now, can be re-opened when needed.


scubamuc (approved):

pachulo:
Thanks for this, but let's wait till the conversation in https://github.com/nextcloud-snap/nextcloud-snap/issues/2456 finishes.

pachulo:
> Not sure if I should remove the script completely or keep as leftover. 

I think that we can remove it completely...WDYT @kyrofa ?

kyrofa:
Yeah let's go ahead and completely remove that file, please.

Pilzinsel64:
I did remove the file now and cleared the history (myself, for the first time ever, yay). 😃 
This is ready to review now!

pachulo (approved):

Issue #2478, opened 2023-07-28
Builds failing due Chrome


Pilzinsel64:
#### What dependency/component is out of date?
Looks like the link to Chrome is not available anymore. See the build errors:
https://github.com/nextcloud-snap/nextcloud-snap/actions/runs/5685667058/job/15411020535#step:8:82

This let fail all the builds at the moment.

Not sure where to get a new link for and how to change in the code. Anyone an idea?

kyrofa:
Yeah, chrome broke all sorts of workflows with their latest shenanigans. We need to move to "chrome for testing". See [this blog post](https://developer.chrome.com/blog/chrome-for-testing/).

Pilzinsel64:
I'm not sure, but looks like I'm a little bit lost here.
![grafik](https://github.com/nextcloud-snap/nextcloud-snap/assets/23138465/a4bbf58b-de8f-4f6d-aec1-47d6d465c563)

Feel free to take over. Branche is `fix/ci-chrome`. 😅

faibz:
Just a guess, but I think the issue might be caused by the `webdrivers` gem. The [GitHub page for it](https://github.com/titusfortner/webdrivers#update-future-of-this-project) seems to mention Chrome breaking post 115, and that updating to Selenium 4.10 ([this?](https://rubygems.org/gems/selenium-webdriver/versions/4.10.0)) should work with Chrome 115. 

If that's correct, then perhaps the `webdrivers` entry should be dropped from `Gemfile.lock` and the `selenium-webdriver` version updated to 4.10.0.

Pilzinsel64:
@faibz Good hint, thanks! Let's try it.
I made a commit just updating the version numbers a bit. Let's see if even this will work.

Pilzinsel64:
Ok, I now tried to removing the `Gemfile.lock` and re-run `bundle install`.

Using the branche `fix/ci-webdirvers` just for testing the testing now. :D
If it success before the year ends, I'll create fresh branches & PRs for `develop`, `26` and `25`. ;)

faibz:
I opened [this PR](https://github.com/nextcloud-snap/nextcloud-snap/pull/2479) which works locally, but is now out of date and has conflicts with the latest version of this fix/ci-chrome branch. Perhaps what's building now will work too as it seems to update the right things (capybara and selenium-webdriver).

Pilzinsel64:
@faibz Thanks a lot!

[...]

PR #2481, opened 2023-07-31
tests: support newer chrome version

faibz:
#2478


kyrofa (changes requested):
Thank you, @faibz. This looks good to go, but before we ship it, would you mind bringing this branch up-to-date with develop and squashing it into a single commit, please?

faibz:
So I guess I'm not great at the whole squashing thing. I squashed all the commits into one but that meant I had a different commit with all the develop changes (and 8 files changed), so needed to update again. Now there are 2 commits but only 4 files changed, which I think would have been the same output had I only squashed commits relevant to the issue.

Is that fine or should I do something else?

kyrofa:
@faibz I took the liberty of squashing completely for you. Mind double-checking? Look okay?

faibz:
Ah, you're a hero. Yeah looks good to me!

kyrofa (approved):

Pilzinsel64 (approved):

Pilzinsel64:
@faibz Do you want to re-make the change with PR for branche `26` abd `25` too? Otherwise I would do that quickly tomorrow morning. :)

PR #2509, opened 2023-08-28
mysql: revert commit be8348a7c3

pachulo:
Here I'm trying to apply the same solution that [the Ubuntu package used](https://bugs.launchpad.net/ubuntu/+source/mysql-8.0/+bug/2019203/comments/16) to fix a bug similar to [that observed by our users](https://github.com/nextcloud-snap/nextcloud-snap/issues/2405).
Their solution was to revert [commit be8348a7](https://github.com/mysql/mysql-server/commit/be8348a7c3e8510b998a063065b626a459631b32), so we will see if that also works for us.

In theory, a proper fix should be coming in the next mysql version ([8.0.35](https://bugs.mysql.com/bug.php?id=110752)), but we will also need to try that, as they only mention aarch64-based platforms.

We'll need to make sure we backport this fix to v25, which is where this issue was introduced.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -330,11 +330,14 @@ parts:
 
   mysql:
     plugin: cmake
-    after: [boost]
+    after: [boost, patches]
 
     # Get from https://dev.mysql.com/downloads/mysql/
     source: https://dev.mysql.com/get/Downloads/MySQL-8.0/mysql-boost-8.0.34.tar.gz
     source-checksum: md5/c8cfab52fbde1cca55accb3113c235eb
+    override-pull: |
+      snapcraftctl pull
+      patch -p1 -d $SNAPCRAFT_PART_SRC < $SNAPCRAFT_STAGE/mysql-revert-BUG-34849343-Aligned_atomic-not-working-as-in.patch
     configflags:
       - -DCMAKE_INSTALL_PREFIX=/
       - -DBUILD_CONFIG=mysql_release
--- a/src/patches/mysql-revert-BUG-34849343-Aligned_atomic-not-working-as-in.patch
+++ b/src/patches/mysql-revert-BUG-34849343-Aligned_atomic-not-working-as-in.patch
@@ -1,0 +1,379 @@
+From 3ee9c37cf3a0a824c2301d355b477c600908338f Mon Sep 17 00:00:00 2001
+From: Kyle Fazzari <kyrofa@ubuntu.com>
+Date: Mon, 18 Sep 2023 17:53:05 -0700
+Subject: [PATCH] Revert "BUG#34849343 Aligned_atomic not working as intended -
+ aligned_alloc"
+
+This reverts commit be8348a7c3e8510b998a063065b626a459631b32.
+---
+ include/my_aligned_malloc.h               | 55 ----------------
+ mysys/CMakeLists.txt                      |  1 -
+ mysys/my_aligned_malloc.cc                | 79 -----------------------
+ sql/memory/aligned_atomic.h               | 71 ++++++--------------
+ unittest/gunit/memory/aligned_atomic-t.cc | 34 ----------
+ 5 files changed, 18 insertions(+), 222 deletions(-)
+ delete mode 100644 include/my_aligned_malloc.h
+ delete mode 100644 mysys/my_aligned_malloc.cc
+
+diff --git a/include/my_aligned_malloc.h b/include/my_aligned_malloc.h
+deleted file mode 100644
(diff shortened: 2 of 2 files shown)


kyrofa:
@pachulo I think this should flow the same direction as every other fix (develop -> backports), but what you've got here is in a good state for testing, so let's leave it this way until we get some proof that it works. Once the tests are green I'll release builds for all architectures to a temporary channel.

pachulo:
@kyrofa can we have a channel created for other folks, apart from @caseydiers , to test this?
Thanks!

scubamuc (approved):

kyrofa:
@pachulo I'm sorry for the delay, these are building now.

kyrofa:
Alright, these have been build for all architectures and released to the `latest/beta/pr-2509` channel.

kyrofa:
Given @jdstrand's [successful test](https://github.com/nextcloud-snap/nextcloud-snap/issues/2405#issuecomment-1722571909) of this PR, I think it's time to clean it up and target the right branch so we can properly review it and backport it. @pachulo you up for that?

pachulo:
Check https://github.com/nextcloud-snap/nextcloud-snap/pull/2530

kyrofa:
@pachulo we already have a branch for testing with this PR's number on it, you should be able to just edit this PR and target it to the proper branch, no?

pachulo:
Well, when I created this I branched from the 25 branch, so I don't really know if that would work...

kyrofa:
Yeah just retarget the PR and rebase the code.

pachulo:
Bufff, I don't really know how to do it, my git-foo is not that strong... :sweat:

kyrofa:
Alternatively, just force-push your develop-revert-mysql-be8348a7c3 branch to this one.

pachulo:
Wow, that was neat, thanks a lot for all the knowledge!

kyrofa:
Yeah perfect, nice work.

snap/snapcraft.yaml
```diff
@@ -330,11 +330,18 @@ parts:
...
     # Get from https://dev.mysql.com/downloads/mysql/
     source: https://dev.mysql.com/get/Downloads/MySQL-8.0/mysql-boost-8.0.34.tar.gz
     source-checksum: md5/c8cfab52fbde1cca55accb3113c235eb
+    override-pull: |
+      snapcraftctl pull
+      patch -p1 $SNAPCRAFT_PART_SRC/mysys/CMakeLists.txt < $SNAPCRAFT_STAGE/mysql-mysys_CMakeLists.txt.patch
+      patch -p1 $SNAPCRAFT_PART_SRC/sql/memory/aligned_atomic.h < $SNAPCRAFT_STAGE/mysql-sql_memory_aligned_atomic.h.patch
+      patch -p1 $SNAPCRAFT_PART_SRC/unittest/gunit/memory/aligned_atomic-t.cc < $SNAPCRAFT_STAGE/mysql-unittest_gunit_memory_aligned_atomic-t.cc.patch
+      rm $SNAPCRAFT_PART_SRC/include/my_aligned_malloc.h
+      rm $SNAPCRAFT_PART_SRC/mysys/my_aligned_malloc.cc
```
kyrofa, line 344: You mention that this is a revert of a [single commit in upstream mysql](https://github.com/mysql/mysql-server/commit/be8348a7c3e8510b998a063065b626a459631b32), but here we're making a number of changes manually. Is there a reason we can't use a single patch file for this?

pachulo: Yes there is: I found out that the files on the `mysql-server` repo at `mysql-8.0.34` tag differs a little bit from the files at mysql-boost-8.0.34, so I had to "adapt" the patch, but it worked OK.

kyrofa: Ah, okay. I think ideally this would be a single patch with some details in it, like a commit message. Let me see if I can help.

kyrofa: Alright, see latest commit

pachulo: Much nicer! How did you manage to do it in the end?

kyrofa: Good deal. I just cloned mysql, did the revert in git, and then used `git format-patch` to get the .patch file.

kyrofa (approved):

snap/snapcraft.yaml
pachulo: And you didn't have to adapt anything for it to apply neatly on the decompressed `mysql-boost-8.0.34.tar.gz`?
I must have missed something then...

PR #2511, opened 2023-09-01
Release 25.0.10snap1

pachulo:


Pilzinsel64 (approved):

scubamuc (approved):

pachulo:
Well, I got it wrong again: this time I created and pushed the git tag to the launchpad repo, but it still built it as `25.0.9snap1+git4.0053e1c`...

pachulo:
Hi @kyrofa ! Can you sping new builds for 25 in LP so they get the correct name from the already pushed tag?
I don't seem to have permissions to do so...

kyrofa:
@pachulo the tags need to be annotated. The versions come from a snapcraft-specific rendition of running `git describe`. If you don't use annotated tags, git doesn't pick them up by default. I fixed your tag, run `git show 25.0.10snap1` to see what I mean. Annotated tags give you a place to basically provide a custom snippet of changelog; you can see the pattern I follow.

I just triggered new builds. You're right, looks like I haven't been disciplined about owners for these-- I've been creating them under my own account! That has now been fixed.

PR #2512, opened 2023-09-02
Release 24.0.12snap1

pachulo:
I guess nobody is using this by now, but let's finish 24.x with style.


scubamuc (approved):

Pilzinsel64 (approved):

Pilzinsel64:
@pachulo Try creating a tag for the commit before merging the PR. Maybe it get pushed too fast and the build at launchpad get started before you pushed the tag. (Just a guess)

kyrofa:
Man, I thought 24 was done!

> Try creating a tag for the commit before merging the PR.

No, don't do that-- the actual release needs to be the merge commit, not THIS commit. That said:

> Maybe it get pushed too fast and the build at launchpad get started before you pushed the tag. (Just a guess)

It's true, once you merge this, you need to quickly push a tag or LP will fire up a build without it. It takes a few minutes for LP to spin a build though, so it's not a tremendous race condition.

Pilzinsel64:
> It's true, once you merge this, you need to quickly push a tag or LP will fire up a build without it. It takes a few minutes for LP to spin a build though, so it's not a tremendous race condition.

@kyrofa Aaah. Good to know. But I assume the tag needs to be pushed manually by someone with right access or that also done automatically? If first, might it be an idea to add a simple delay of a few minutes (maybe `sleep 300`) before the workflow pushes the changes?

PR #2537, opened 2023-09-21
Release 27.1.1snap1

kyrofa:
- mysql: revert BUG#34849343 Aligned_atomic not working as intended
- nextcloud: update to 27.1.1


.github/workflows/launchpad.yml
```diff
@@ -2,11 +2,13 @@ name: Deploy to Launchpad
...
 on:
   push:
-    # Only applies to our protected branches
+    # Only applies to our protected branches and tags
     branches:
       - master
       - develop
       - '[0-9]+'
+    tags:
+      - '*'
```
Pilzinsel64, line 11: Uh nice! That will push to launchpad also when we create a tag for the previously created merge commit, right? :)

Pilzinsel64 (approved):
👍

.github/workflows/launchpad.yml
kyrofa: Yes that's correct.

kyrofa: Agh, this doesn't actually work due to https://github.com/github/roadmap/issues/825 .

PR #2542, opened 2023-09-25
initial port to core22

soumyaDghosh:


pachulo:
Hi! First, thanks for your contribution @soumyaDghosh !

I'm afraid this is will need some more work: we make use of [Snapcraft local plugins](https://snapcraft.io/docs/writing-local-plugins), and those need to be adapted when upgrading the core used by the snap.
Besides, I think they are deprecated on `core22`, not on Snapcraft 7 as the documentation says, so we must:
1. upgrade the local plugins to `core20`
2. and then wait for Snapcraft's alternative for custom plugins in...`core24`?

Hope that helps!

kyrofa:
> If you can simply those in a shell script, that can help me.

If that was trivial, it's what we would have done in the first place. Doing it that way will require some non-trivial refactoring. Not sure why snapcraft decided to go this way.

pachulo:
> Not sure why snapcraft decided to go this way.

From what I read [here](https://forum.snapcraft.io/t/local-v2-plugin-base-core22/32227/4):
> to prevent possible issues with e.g. Python interpreter compatibility, or to avoid exposing too much of the internals to the external plugin implementation.

Maybe we can describe our usage of local plugins in that thread, to see if it makes sense to have them "upstreamed" on https://github.com/canonical/craft-parts

kyrofa:
> see if it makes sense to have them "upstreamed" on https://github.com/canonical/craft-parts

No thanks, I've been on the other side of that. I don't want to hand someone else the ability to break us any more than they already can.

pachulo:
Hi again @soumyaDghosh ! We want to try to lift the snap to use core22 and all the help we can get will be greatly appreciated.
All the information about how we should refactor the local plugins into something else is documented [here](https://github.com/nextcloud-snap/nextcloud-snap/issues/2908).

Would you be interested in participating?

soumyaDghosh:
I actually have all the parts ported to the new plugin system. But when I was working on it, there was some issue with some source of a part (prolly reddis), so, I left it and then I forgot about it.

pachulo:
Wow! Can you share the WIP code?

soumyaDghosh:
I tried to build this again yesterday. But sadly, the builder stopped showing no space left. 

https://dpaste.com/6J47BL6LF

soumyaDghosh:
https://launchpadlibrarian.net/778324283/buildlog_snap_ubuntu_jammy_amd64_snapcraft-nextcloud-6e4add5a88be4949fd7df08cf918a57b_BUILDING.txt.gz

This is a build log.

pachulo:
Would you mind opening a new PR with the changes that you made @soumyaDghosh ?

pachulo:
Hope you don't mind that I took your code and opened the PR myself.

PR #2557, opened 2023-10-04
php: add module `pdo_pgsql`

### LINKED ISSUE #2555: migrate from snap to AIO - add `pdo_pgsql` PHP extension.
#### Describe the bug
I need to use Nextcloud Office, which is not supported and does not work anymore on the snap install. 
Therefore I like to migrate my whole install to Nextcloud AIO. In the documentation of AIO its stated by @szaimen that the `pdo_pgsql` PHP extension needs to be added to provide a pathway for migration. (https://github.com/nextcloud/all-in-one/commit/c54395aa4c695fb53e6110328ede473c13693de2)

feature request :)

Pilzinsel64:
Fix #2555

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -251,6 +251,7 @@ parts:
       - libgmp-dev
       - libzip-dev
       - libargon2-0-dev
+      - libpq-dev
 
       # This is no longer bundled with PHP as of v7.4
       - libonig-dev
@@ -278,6 +279,7 @@ parts:
       - libzip4
       - libargon2-0
       - libonig4
+      - libpq5
     prime:
      - -sbin/
      - -etc/
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -234,7 +234,7 @@ parts:
       - --with-password-argon2
 
       # Enable pdo_pgsql
-      - --with-pgsql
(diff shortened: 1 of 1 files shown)


scubamuc (approved):

pachulo:
I guess it's not jut a matter of adding the php compile option; look at the error @Pilzinsel64 :

> configure: error: Cannot find libpq-fe.h. Please specify correct PostgreSQL installation path

pachulo:
> But I think we don't really want to ship a mostly useless database installation. Or what do you think?

If that's the case, totally. But what I understood from [here](https://github.com/nextcloud/all-in-one/blob/main/migration.md#migrate-the-files-and-the-database) is that only the PHP extension is needed, not a whole PSQL instance.

Pilzinsel64:
But what's the way to go to install that extension in snap? I thought the --with parameter is the way to go.

Can we install packages via apt-get or do wee need to get the archive and build from source? Or is there something else I missed? 🤔

pachulo:
Let's see if it compiles with the packages.

kyrofa:
I mentioned this on the issue but I'll put it here as well: [I recommend not doing this](https://github.com/nextcloud-snap/nextcloud-snap/issues/2555#issuecomment-1747898747).

PR #2560, opened 2023-10-05
nextcloud: update to 27.1.2

### LINKED ISSUE #2559: Update nextcloud to 27.1.2
https://nextcloud.com/changelog/

Pilzinsel64:
Fix #2559

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -181,8 +181,8 @@ parts:
 
   nextcloud:
     plugin: dump
-    source: https://download.nextcloud.com/server/releases/nextcloud-27.1.1.tar.bz2
-    source-checksum: sha256/3a91500566874675676fa3b5bfae2587a839cde41dfac5318043b162c1311fab
+    source: https://download.nextcloud.com/server/releases/nextcloud-27.1.2.tar.bz2
+    source-checksum: sha256/0742b247aaee0b7044db0062f0a914aa77338c7a7d8fe7da0917147d76689721
     organize:
       '*': htdocs/
       '.htaccess': htdocs/.htaccess


stondino00 (approved):

stondino00:
This is failing to build in launchpad it appears @kyrofa . Thanks

pachulo:
I've just relaunched the builds that failed (armhf was succesful)

pachulo:
I don't know what's going on, arm64 build is failing with this error:

```
Running repo phase...
Cloning into 'nextcloud'...
fatal: unable to access 'https://git.launchpad.net/nextcloud-snap/': gnutls_handshake() failed: The TLS connection was non-properly terminated.
Build failed
```
The rest of the builds went OK and are on the `latest/beta` channel.
I've tried to rebuild it a few time now...any ideas @kyrofa ?

Pilzinsel64:
But the build on PF went fine, why? 🤔

kyrofa:
Looks like a temporary outage is all. Builds look green again.

scubamuc:
@Pilzinsel64 i'll need the machine for testing fresh install... are you finished there? may I kill your install?

Pilzinsel64:
@scubamuc Of course, feel free.

scubamuc:
Fresh install tests completed without issues:

grafik

grafik

scubamuc:
The Logs for `latest/beta` are showing the FQDN-warning again after fresh install. @Pilzinsel64, so we were mistaken thinking that this was gone since mDNS was removed. We'll need to figure out "setting FQDN" vs. "CODE and Office" `/etc/hosts`using localhost only:

```
Oct 09 13:50:58 NC-SNAP--testing nextcloud.apache[286960]: AH00558: httpd: Could not reliably determine the server's fully qualified domain name, using 192.168.2.222. Set the 'ServerName' directive globally to suppress this message
```
@kyrofa, do you know another method to set the 'ServerName' directive other that `/etc/hosts`. I was thinking along the lines of `hostnamectl`, but the is no Option for the FQDN. According to [this](https://yewtu.be/watch?v=bA8GVdGFT30) setting FQDN is "best practice".

Pilzinsel64:
The only important thing to add:
The nextcloud server needs to be able to make https requests to it's FQDN, so it need to go throw the proxy server. This is how nextcould want to work.
Editing the hosts file does work if the nextckoud server also handles https. But if it doesn't because there is another server as proxy between, then there is another way needed I guess.

Pilzinsel64:
The only important thing to add:
The nextcloud server needs to be able to make https requests to it's FQDN, so it need to go throw the proxy server. This is how nextcould want to work.
Editing the hosts file does work if the nextckoud server also handles https. But if it doesn't because there is another server as proxy between, then there is another way needed I guess.

I have read that some people said adding "ServerName localhost" to https.conf would be enough. 🤔 

@scubamuc Isn't 192.168.2.222 your proxy? Did you try adding an A record to your DNS (if you have one) with your FQDN and the IP of your proxy as destination? Or as alternative, you can try adding an entry to the /etc/hosts file to your nextckoud server with the IP of your proxy and the FQDN.

scubamuc:
@Pilzinsel64 thanks for pointing that out. nope x.x.x.222 is not my proxy, but I've added:
```
<x.x.x.ProxyIP> <my.sexy.domain.io> <servername>
```
to `/etc/hosts` and FQDN warning has disappeared.

Do you think I should add that to Wiki?

PR #2568, opened 2023-10-17
ci/cd: tune ext4 for speed

pachulo:
I want to test if builds are quicker with this change.

Inspired by https://github.com/canonical/lxd/pull/12385

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -33,6 +33,19 @@ jobs:
     runs-on: ubuntu-20.04
 
     steps:
+      # Inspired by https://github.com/canonical/lxd/pull/12385
+      - name: Performance tuning
+        run: |
+          set -eux
+          # optimize ext4 FSes for performance, not reliability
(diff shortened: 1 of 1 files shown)


scubamuc (approved):

pachulo:
Well, it seems like there was some improvement in build tme: from [1:58 hours](https://github.com/nextcloud-snap/nextcloud-snap/actions/runs/6552311695/job/17795346661?pr=2567) to [1:38](https://github.com/nextcloud-snap/nextcloud-snap/actions/runs/6552636162/job/17796363722?pr=2568) or so.

I would say it's worth the change, WDYT @kyrofa ?

kyrofa:
That's a small enough change I wouldn't be surprised if it was due to less of a runner load at the time or something. I recommend calculating averages: run this PR's integration suite 5 times, and run master's 5 times, and take the average of each. Each run on master should happen at roughly the same time of day as this PR, so we can ignore the load differences.

Issue #902, opened 2019-02-19
ResourceLocator can not find a web root on several extensions


asadana:
### Server configuration

**Operating system**: Ubuntu 18.04.2

**Nextcloud version:** 15.0.2

**Updated from an older Nextcloud/ownCloud or fresh install:** Upgraded from 14.0.6

**Where did you install Nextcloud from:** Snappy

</details>

**Are you using external storage, if yes which one:** Using S3 as primary storage

**Are you using encryption:** no

</details>

### Client configuration
**Browser:** Mozilla Firefox 65 and Google Chrome 72

**Operating system:** Linux (Ubuntu/Fedora) and Windows 10

#### Nextcloud error logs
<details>
<summary>Error log: Talk/Spreed</summary>

```
[lib] Error: ResourceLocator can not find a web root (root: /var/snap/nextcloud/11336/nextcloud/extra-apps/spreed, file: index.php/css/spreed/656a-c5cf-merged-files.css?v=f7fbce5eca1c766bdf8f93c3337c21ac, webRoot: , throw: true)

GET /index.php/apps/files/
from  200.00.200.200 by xxxx@xxx.com at 2019-02-18T16:11:30-05:00
```
</details>

<details>
<summary>Error log: Calendar</summary>

```
[lib] Error: ResourceLocator can not find a web root (root: /var/snap/nextcloud/11336/nextcloud/extra-apps/calendar, file: index.php/css/calendar/029b-c5cf-app.min.css?v=f7fbce5eca1c766bdf8f93c3337c21ac, webRoot: , throw: true)

GET /index.php/apps/calendar/
from  200.00.200.200 by yyyy@yyy.com at 2019-02-18T14:17:03-05:00
```
</details>

<details>
<summary>Error log: Deck</summary>

```
[lib] Error: ResourceLocator can not find a web root (root: /var/snap/nextcloud/11336/nextcloud/extra-apps/deck, file: index.php/css/deck/7822-c5cf-style.css?v=f7fbce5eca1c766bdf8f93c3337c21ac, webRoot: , throw: true)

GET /index.php/apps/deck/
from 200.00.200.200 by zzzz@zzz.com at 2019-02-18T13:50:15-05:00
```
</details>

asadana:
Possible duplicate of #625.

Spreed/Talk has had much more frequent errors than other two. Basically any use of Talk spawns these errors.
Calendar and deck errors are more sparsely seen.

kyrofa:
[...]

PR #2626, opened 2023-12-07
Enable dailies for Nextcloud 28

Pilzinsel64:
- Enable dailies for Nextcloud 28, as they are already available.
- Disable dailies for Nextcloud 25, as this version is EOL.

### DIFF
--- a/.github/spawn-dailies.sh
+++ b/.github/spawn-dailies.sh
@@ -1,7 +1,6 @@
 #!/bin/bash
 
 latest_master_url="https://download.nextcloud.com/server/daily/latest-master.tar.bz2"
-latest_stable25_url="https://download.nextcloud.com/server/daily/latest-stable25.tar.bz2"
 latest_stable26_url="https://download.nextcloud.com/server/daily/latest-stable26.tar.bz2"
 latest_stable27_url="https://download.nextcloud.com/server/daily/latest-stable27.tar.bz2"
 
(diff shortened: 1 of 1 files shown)


scubamuc (approved):

pachulo (approved):

Pilzinsel64:
Hm, the dailies will be uploaded to launchpad but the channel `28/edge` is not available. Do I miss something? 🤔

Pilzinsel64:
I tested a local test-install of `28/edge`. basically everything works as expected. I also tried to update my private production isntance to `28/edge` and here too, basically it works fine. But I got around a bit trouble and finally reverted back to 27.

The issue that is relevant to snap and prevented me from keep tracking 28/edge on my private prod instance for long-term-testing were a single long-flooting issue. See this pic for the errors (recorded on the test instance).
grafik

On my private prod instance the errors has a few more such errors. That's because I have a huge number of additional apps installed.

Note: On my both other prod instances (both manually installed) this erorrs do not exists even with the same apps and versions installed.

However, I found this old issue here: https://github.com/nextcloud-snap/nextcloud-snap/issues/902
@kyrofa @pachulo What do you think? Should I open a new issue there or comment on the existing? Or is that something on our side?

Pilzinsel64:
Also, as I read the issue above, it might not be breaking anything, or? Can I ignore that a just log-flooting bug and update again my private prod to 28/edge for futher testing until next stable release or do you recommend me to stay on 27 until we solved that issue?

kyrofa:
We haven't changed anything in the config-- this sounds like an upstream Nextcloud issue again. I do recommend logging an issue, please.

Pilzinsel64:
I made one, see https://github.com/nextcloud/server/issues/42671

Pilzinsel64:
I see that also AIO stable waits for this / a similar issue: https://github.com/nextcloud/all-in-one/issues/3888 https://github.com/nextcloud/server/issues/42493

Issue #1075, opened 2019-07-25
common name in self-signed certificate


Smoin1:
(please excuse any mistakes as this is my first entry ever here. I just created an account ;) )
#### Describe the bug
Several systems (e.g. Linux machines via mount command) cannot use nextcloud data because common name attribute is missing in self-signed certificate 

#### To Reproduce
1. Setup nextcloud via snap
2. enable self-signed certificate
3. Mount share via linux

#### Expected behavior
Add common name attribute (and potentialle SubjectAltNames too) to certificate to prevent those errors from happening.

#### OS/snapd/snap version
What OS and version are you running? Please also include the output of the
following commands:

    $ snap list nextcloud
Name       Version       Rev    Tracking  Publisher   Notes
nextcloud  15.0.10snap2  14827  stable    nextcloud✓  -

and:

    $ snap version
snap    2.39.3
snapd   2.39.3
series  16
ubuntu  18.10
kernel  4.18.0-17-generic

#### Logs
Please paste in the output from
[this script](https://github.com/nextcloud/nextcloud-snap/wiki/Issue-Debugging-Scripts)
to assist in debugging (note that it's pre-formatted).

kyrofa:
> Several systems (e.g. Linux machines via mount command) cannot use nextcloud data because common name attribute is missing in self-signed certificate

I have no problems mounting davs via Nautilus, for example. I of course needed to click the "yes yes, trust the darn thing" prompt like all other self-signed certs. How are you trying to use the `mount` command? Are you sure it's not failing just because it's self-signed? I'd be surprised if the CN was required.

kyrofa:
Closing due to inactivity.

Smoin1:
I am sorry, I missed the answer!

Smoin1:
![image](https://user-images.githubusercontent.com/53152348/72462277-fe749580-37d0-11ea-81f8-d8058c63e881.png)

Smoin1:
btw: this is just one of multiple machines where I had this problem. also some Browser and mobile devices did not like the missing attribute

Smoin1:
An example from brwoser in Windows: 
![image](https://user-images.githubusercontent.com/53152348/72462845-5c55ad00-37d2-11ea-9bfe-c17d8e3713b9.png)

thalesmaoa:
Just having the same problem. Is there a walkaround?

kyrofa:
[...]

PR #2639, opened 2023-12-28
Update https-utilities

dqos:
Add CommonName to self-signed certificate to fix: https://github.com/nextcloud-snap/nextcloud-snap/issues/1075

### DIFF
--- a/src/https/utilities/https-utilities
+++ b/src/https/utilities/https-utilities
@@ -36,7 +36,7 @@ generate_self_signed_certificate()
 	chmod 750 "$(dirname "$SELF_SIGNED_CHAIN")"
 
 	openssl req -newkey rsa:4096 -nodes -keyout "$SELF_SIGNED_KEY" \
-	        -x509 -days 90 -out "$SELF_SIGNED_CERT" -subj "/O=Nextcloud/CN=Nextcloud"
+	        -x509 -days 90 -out "$SELF_SIGNED_CERT" -subj "/O=Nextcloud/CN=$(hostname -f)"
 
 	rm -f "$SELF_SIGNED_CHAIN"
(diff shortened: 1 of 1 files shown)


src/https/utilities/https-utilities
```diff
@@ -36,7 +36,7 @@ generate_self_signed_certificate()
 	chmod 750 "$(dirname "$SELF_SIGNED_CHAIN")"
 
 	openssl req -newkey rsa:4096 -nodes -keyout "$SELF_SIGNED_KEY" \
-	        -x509 -days 90 -out "$SELF_SIGNED_CERT" -subj "/O=Nextcloud"
+	        -x509 -days 90 -out "$SELF_SIGNED_CERT" -subj "/O=Nextcloud/CN=Nextcloud"
```
kyrofa, line 39: You literally only need a CN entry? It doesn't even matter what is is?! This seems broken: the CN is supposed to be the domain name you're using to visit the site protected by the cert.

dqos: Yes, this fixes it in order to mount webdav drives. For example davfs2 complaints about a missing CN:
```
/sbin/mount.davfs: Mounting failed.
Server certificate was missing commonName attribute in subject name```

The CN value is actually deprecated, browsers use the SAN value. But for some reason, davfs2 checks the CN. I think in order to do it properly, we can fill in the current hostname of the server.

dqos: I have adjusted the CN to be dynamic now based on the hostname. Feel free to merge or close, it fixed it for our completely internal setup 😄

scubamuc (approved):

kyrofa (changes requested):
I don't think we want this as it stands today. If we want to make the CN configurable, we can do that, but it'll be a larger feature through the config hook. If you want to take a crack at it I can help guide you.

PR #2642, opened 2024-01-10
php: update to 8.2.14

### LINKED ISSUE #2564: Update PHP to 8.2
#### What dependency/component is out of date?
This is not personally required, or related to a particular security issue.

I noticed that the recommended version by the Nextcloud docs was [8.2](https://docs.nextcloud.com/server/stable/admin_manual/installation/system_requirements.html) whilst doing some maintenance, although 8.1 is still acceptable.

Pilzinsel64:
fix #2564

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -195,8 +195,8 @@ parts:
 
   php:
     plugin: php
-    source: https://php.net/get/php-8.1.25.tar.bz2/from/this/mirror
-    source-checksum: sha256/a86a88c1840c1bc832bcfd2fbec3b8a1942c8314da5dff53f09f9c98d0c12e8a
+    source: https://php.net/get/php-8.2.14.tar.bz2/from/this/mirror
+    source-checksum: sha256/f871e131333d60ae6c537b1adddbc2aea54c436c562af986fb8309c060040b9e
     source-type: tar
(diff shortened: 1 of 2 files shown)


scubamuc (approved):

Pilzinsel64:
I don't really like that missing redis extension: https://stackoverflow.com/questions/76561932/why-is-the-redis-extension-not-working-with-php-8-2

pachulo:
Hi @Pilzinsel64 ! I think you need to update more things, apart from the snapcraft.yml.
Check the [last PR with an upgraded branch of PHP from](https://github.com/nextcloud-snap/nextcloud-snap/pull/2318) @kyrofa

Pilzinsel64:
Alright, so I'll close this for now. I have no idea what's up with that redis module and I would need a lot try-and-error (and time) to find it out. Probably @kyrofa knows what to do, so I'll let this up for him. :)

kyrofa:
Given that we build the redis extension from scratch, that may not actually be a problem @Pilzinsel64. As the others have said though, when bumping PHP's minor version, there are other changes you need to make, or PHP won't run. That's your problem at this point anyway. If we get that config proper, then we can see if the redis extension has issues or not.

kyrofa:
Took my own crack at this in #2654.

PR #2650, opened 2024-01-16
Upgrade MySQL to 8.0.36

### LINKED ISSUE #2649: Upgrade MySQL to 8.0.36
Oracle Q1 Security Updates Today

https://www.oracle.com/security-alerts/cpujan2024.html

stondino00:
Fixes #2649

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -333,8 +333,8 @@ parts:
     after: [boost]
 
     # Get from https://dev.mysql.com/downloads/mysql/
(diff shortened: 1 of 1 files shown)


Pilzinsel64 (changes requested):
Looks like the md5 checksum looks wrong. 🤔

stondino00:
Maybe it's too early in the am for me but it's showing wrong md5 for the boost package which isn't changed in this commit.  I'm showing all check-sums as correct in this commit.  Let me know if I'm missing something or it just needs re-ran.

stondino00:
Ok looks like re-running it is working like expected now.

Pilzinsel64 (approved):

Issue #2641, opened 2024-01-10
Update Nextcloud to 28


Pilzinsel64:
### Pending todos
- [x] Wait for 28.0.4 stable
- [x] https://github.com/nextcloud-snap/nextcloud-snap/pull/2665
- [x] Create branch `27`
- [x] Create label `27`

### Pending PRs
- [x] ~~https://github.com/nextcloud-snap/nextcloud-snap/pull/2657~~
- [x] https://github.com/nextcloud-snap/nextcloud-snap/pull/2655
- [x] https://github.com/nextcloud-snap/nextcloud-snap/pull/2659

### Pending upstream issues
- [x] https://github.com/nextcloud/server/issues/42286
- [x] https://github.com/nextcloud/server/issues/42671

Pilzinsel64:
@kyrofa @pachulo I would like to assign the tasks about creation of branch `27` and merging of #2665 to you, if you agree and have that time. I was thinking about creating the 27 branch myself to be able to merge and test #2668 but I have no idea what needs to be done at launchpad regarding the 27 branche and even don't have any abilities to do that. So I probably should leave that to you before I break anything. ;)

kyrofa:
Done.

Pilzinsel64:
Alright, after a quick test on my prod I found out that the log-flooting bug still exists. However, this does NOT happen as long as you just install the raw snap package without any additional or recommended apps.

See https://github.com/nextcloud/server/issues/42671#issuecomment-1925888315

I recommend to not move this to stable until we (or Nextcloud) found a solution as many server owner do probably additional apps.

In my case there are 8 apps that cause that error. But for sure, this is a problem in the core `server` repo and NOT of the apps.

Everything else does work fine as far as I can tell after 30 minutes testing. If there is something else, I will let you know here.

Pilzinsel64:
I suggest to #2659 for now and just push a stable 28 release.

The maintenance window config is new and the warning too but the behavior how background jobs are handled is still the same then NC27 until you configure a maintenance window. The other problem is that the config uses UTC format and shell specs are still missing - wich means the PR needs more work anyway.

NC28 on `latest/beta` seems to be very stable otherwise. I would make a release PR later this day.

What do you think?

scubamuc:
CODE & Nextcloud Office test:
[...]

PR #2655, opened 2024-01-19
php: enable sodium

Pilzinsel64:
This resolved the nice warning in NC28 by installing the php sodium extension. 

Fix #2643
Required for #2641 

grafik

It is a recommended php module by Nextcloud:
> PHP module sodium (for Argon2 for password hashing. bcrypt is used as fallback, but if passwords were hashed with Argon2 already and the module is missing, your users can’t log in.)

PR #2657, opened 2024-01-19
apache: enable `application/javascript` mimetype

### LINKED ISSUE #2656: [NC28] Missing `application/javascript` mimetype
Nextcloud 28 complains about missing Javascript mimetype. We should fix that before releasing NC28 packages.
grafik

Pilzinsel64:
Fix #2656
Required for #2641 

Maybe also fix the ResourceLoader webroot issue! Need to be tested when dailies have this fix included.


Pilzinsel64:
It's so sad, this change needs to go to master until I can verify if it's working. Same for #2655

scubamuc (approved):

kyrofa:
Yes, it appears that apache ships its own:

> Failed to stage: Parts 'apache' and 'apache-customizations' have the following files, but with different contents:
>     conf/mime.types

Pilzinsel64:
Do you know any way to pass this? Maybe without coding?
This change seems to be important for NC28+.

kyrofa:
Not off the top of my head. We could always patch that file when building Apache if we need to.

Pilzinsel64:
Oh dear, we probably don't need this. See https://github.com/nextcloud/server/issues/42989 Fix in https://github.com/nextcloud/server/pull/42999

When I tested nc28 on my prod instance two weeks back I didn't have this error. Also when I write the hostname to the config, it works. I am also able to download the file `https://ncsnaptest/apps/settings/js/esm-test.mjs` as `text/javascript`.

grafik

PR #2659, opened 2024-01-19
nextcloud: add default config for `maintenance_window_start`

### LINKED ISSUE #2658: [NC28] Configure maintenance time for daily resource intensive background jobs
grafik

Config to add ("1" stands for the first hour of the day, so 1 a.m.):

    'maintenance_window_start' => 1

Ref: https://docs.nextcloud.com/server/28/admin_manual/configuration_server/background_jobs_configuration.html#maintenance-window-start

Pilzinsel64:
Fix #2658
~~Required~~ for #2641

### DIFF
--- a/src/nextcloud/config/config.php
+++ b/src/nextcloud/config/config.php
@@ -55,4 +55,5 @@ $CONFIG = array(
 'log_type' => 'file',
 'logfile' => $snap_data_current.'/logs/nextcloud.log',
 'logfilemode' => 0640,
+'maintenance_window_start' => 1,
 );
--- a/src/nextcloud/fixes/existing-install/7_add_maintenance_window.sh
+++ b/src/nextcloud/fixes/existing-install/7_add_maintenance_window.sh
@@ -1,0 +1,7 @@
+#!/bin/sh -e
+
+# Check if config 'maintenance_window_start' doesn't exist or has no value set
+if ! occ -n config:system:get maintenance_window_start; then
+	# Set default 'maintenance_window_start' to 1 a.m. (UTC)
+	occ -n config:system:set maintenance_window_start --value=1 --type=integer
+fi


Pilzinsel64:
However, I'll test the dailies as soon as this is in master.

scubamuc (approved):

scubamuc:
added to wiki documentation
https://github.com/nextcloud-snap/nextcloud-snap/wiki/Configure-config.php#setting-default-maintenance-window

kyrofa:
It's been a while since we've made a config change, but I'm pretty sure the change you've made here will only effect new installs. Is this something we should be adding for existing installs as well?

Pilzinsel64:
> Is this something we should be adding for existing installs as well?

Yes, of course. Should I put it in autoconfig instead?

kyrofa:
No, that has the same issue. Honestly, I don't think we've _ever_ been in a situation like this, where ideally we'd patch the current config to have a default value, but the user should also be able to override it. The only thing even close is where we [twiddle the redis config](https://github.com/nextcloud-snap/nextcloud-snap/blob/master/src/nextcloud/bin/setup-nextcloud#L46). But we don't care if the user set that to something else, because the snap isn't designed for that to be configurable by the user. In this case... we do. I need to mull this over a bit, but I'm leaning toward using our [migrations capability](https://github.com/nextcloud-snap/nextcloud-snap/tree/master/src/migrations) for this. Thoughts?

Pilzinsel64:
> I'm leaning toward using our [migrations capability](https://github.com/nextcloud-snap/nextcloud-snap/tree/master/src/migrations) for this. Thoughts?

Using that migrations capability seems to be perfect for this! I assume they will be called whenever the snap package version is same or higher then the folder version and then it calls every file in alphabetical order?

kyrofa:
You can see how it works [here](https://github.com/nextcloud-snap/nextcloud-snap/blob/master/src/migrations/bin/run-snap-migrations). It's run via the [post-refresh](https://github.com/nextcloud-snap/nextcloud-snap/blob/master/src/hooks/bin/post-refresh) hook.

As you can see, we only have one migration so far. Let's use it to understand how this works.

In snap version 19.0.3snap3, we changed the logging strategy, logging to completely different files. We didn't want to leave old logs lying around, and we wanted the new logs to include the old data, so as part of installing 19.0.3snap3, we wanted to move the old logs to the new location.

First rule: **if you have a snap version VERSION where you know that an upgrade from anything older has to have something happen to it in order to be compatible with VERSION, create a new directory within the `src/migrations/migrations` directory named the same as VERSION.**

Second rule: **there could be multiple migrations that need to happen in this situation. You can create multiple scripts within the VERSION directory, and they will be run in order (so number them to specify order).**

Then, [run-snap-migrations](https://github.com/nextcloud-snap/nextcloud-snap/blob/master/src/migrations/bin/run-snap-migrations) will compare the version of the snap _before_ it refreshed, OLD_VERSION, to each of the migration directories. It will run the migrations for any directory in there that represents a snap version NEWER than OLD_VERSION.

Does that make sense? So if we know that this is required for snap version `28.0.0snap1`, we create a migration directory named that, and create a new shell script named, I don't know, `1_add-cron-maintenance-window.sh` in there.

My hook knowledge is rusty, but I think I just realized a hairy aspect of this: I'm pretty sure post-refresh runs before services are fired back up, so our ability to use `occ` in here might be pretty limited, depending on how it works (e.g. does it require access to the database?)

Pilzinsel64:
> I'm pretty sure post-refresh runs before services are fired back up, so our ability to use `occ` in here might be pretty limited, depending on how it works (e.g. does it require access to the database?)

While maintenance mode we still could edit configs. But if the services aren't even running, we can only modify the file `config.php` directly.

Pilzinsel64:
I added such a script. Well, the condition is that we publish a release for the upcomming nc28.0.2. But I am confident with that because the errors in the log are not anymore present in that mass, just after ending maintenance mode. As long as we get the other PR with the javascript mime type done, we can beta-test and probably make a release sooner or later. And if needed, we can always rename the folder, if needed.

I can not test the script today, need to do that tomorrow. Feel free to test the script yourself. But it should work™.

src/migrations/migrations/28.0.2snap1/1_add-maintenance-window-start.sh
```diff
@@ -0,0 +1,22 @@
...
+search_text="maintenance_window_start"
+
+# New config text
+insert_text="'maintenance_window_start' => 1,"
+
+# Check if config name exists
+if grep -q "$search_text" "$config_file"; then
+    echo "The text '$search_text' is already available in the config file '$config_file'."
+else
+    # Einfügen des Textes in die vorletzte Zeile der Datei
```
pachulo, line 19: Could this be in English too? My German is a little bit rust... :wink:

Pilzinsel64: Oh, missed that. Done! :)
Will rebase and squash it tomorrow.

src/migrations/migrations/28.0.2snap1/1_add-maintenance-window-start.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
kyrofa, line 1: This warning will be shown as of v28.0.0, right? Rather than trying to forward-guess the snap version we'll release, I recommend just using the version where this is actually needed: `28.0.0snap1`. Yes, that will never exist, but it will make more sense in a year when we wonder why this feature was introduced in v28.0.2 :stuck_out_tongue: .

kyrofa:
> The script works on my prod instance when executing it manually.

I'd like to see it tested for real, though, in the hooks. It will probably require building this locally, are you setup for that? Also, given the risk of sedding config files, I think I'd like to see [shellspec tests](https://github.com/nextcloud-snap/nextcloud-snap/tree/master/tests/unit) for this.

src/migrations/migrations/28.0.2snap1/1_add-maintenance-window-start.sh
Pilzinsel64: Ah, didn't know that it would work. Thought it just checks "equals" and not "equals or higher". 😄

Pilzinsel64:
> I'd like to see it tested for real

Doesn't that requie a package named like "28.0.0snap1" or higher? Even if built locally? (Sorry, don't have enought knowledge to answer that myself.)

> are you setup for that?

I built the snap manually two times in the past. But I always had problems I would need to setup everything again, so I avoid it where I can. 🤐

> I'd like to see [shellspec tests](https://github.com/nextcloud-snap/nextcloud-snap/tree/master/tests/unit) for this.

😳😥

pachulo:
Hi @Pilzinsel64 ! Will you have time to check this now that 28.0.4 has been released?

Thanks!

scubamuc:
@pachulo, don't know if this is relevant, but I did test "set maintenance window" in  #2704 as per [FAQ](https://github.com/nextcloud-snap/nextcloud-snap/wiki/FAQ's#q-how-can-i-set-default-maintenance-window-in-nextcloud-snap) is that not default?... not sure what else needs to be done? 

added this to [testing script](https://github.com/nextcloud-snap/nextcloud-snap/wiki/Develop-and-contribute#setup-testing-environment-per-script-optional) also

Pilzinsel64:
Oh, I am so sorry. I absolutely lost this our of my mind. Also I was werey busy the last weeks.
Tomorrow I'll have some time and I'm going to try to create the shell specs there. :)

Pilzinsel64:
I had a look to shell specs now. Sadly, I really understand how shell specs are done and how I have to make specs for script here. I would welcome if someone else who knows what he does and don't need hours for it could takeover this part. 😟 I'll learn from this then.

Then I was going to test this locally via building from source. Sadly, the wiki entry for this has been gone. To be honest, it was really outdated but it helped be a lot going throw the step of building the snap from source. And I haven't noted the steps on my end in the past. :/

That's why there is another PR now: #2709 I'm not sure yet if that works, but it has a additional commit with an additional tag assigned - based on this branche. It (hopefully) will allow us to test this like other PRs by installing fromt he channel `latest/beta/pr-2659`. Sorry if this route is not welcome. You can be stupid, you just have to know how to help yourself.

After I tested #2659 I'll let you know here. Maybe @scubamuc and someone else can also test. I'm going to test a fresh install from `latest/beta/pr-2659` and then try a upgrade from `latest/stable` to this channel.

scubamuc:
Fresh install tests completed **with** issues

`snap install nextcloud --channel=latest/beta/pr-2659`

grafik

ran usual tests as per [testing environmenmt](https://github.com/nextcloud-snap/nextcloud-snap/wiki/Develop-and-contribute#setup-testing-environment-per-script-optional) **without** setting the maintenance window as per [FAQ](https://github.com/nextcloud-snap/nextcloud-snap/wiki/FAQ's#q-how-can-i-set-default-maintenance-window-in-nextcloud-snap). I'm guessing that was the goal of the test?

grafik

Pilzinsel64:
Oh well, it was a try at least. I didn't know that snap names of beta channels will not include the tag. 🤔 Maybe I'll find another tutorial how to build snaps locally. Can take a look tomorrow again.

Pilzinsel64:
Hey. I'm so soo that I didn't come to this until now. But I'm sure to be able to spend a bit time for on Thursday. If someone else comes to it before, feel free. Sorry again. :/

pachulo:
OK, just did some test with this.
1. I've cloned your repo and checked out this branch:
    * `git clone https://github.com/Pilzinsel64/nextcloud-snap.git && cd nextcloud-snap && git checkout fix/nextcloud/default-config-maintenance-window-start`
1. I then created an annotated tag:
    * `git tag -a 28.0.4snap1`
    * With the following text:
    ```
    Version 28.0.4snap1
    
    - nextcloud: update to 28.0.4
   ```
1. I built it locally using snapcraft 7.x:
    * `sudo snap install snapcraft --channel 7.x/stable && snapcraft build --use-lxd`
1. I then installed nexctloud from the `latest/stable` channel:
    * `sudo snap install nextcloud`
1. And then "uggraded" it using the `.snap` file that was created on step 3:
    * `sudo snap install nextcloud_28.0.4snap1-dirty_arm64.snap --dangerous`

When doing this, I didn't get the `maintenance_window_start` entry added to the `config.php` file.

But then I realized that the script you created didn't have the execution bit set, [but the example @kyrofa shared did](https://github.com/nextcloud-snap/nextcloud-snap/blob/master/src/migrations/migrations/19.0.3snap3/1_move-logs.sh), so I added it with:
`chmod +x src/migrations/migrations/28.0.0snap1/1_add-maintenance-window-start.sh`

After that, I started clean, repeated the steps above and the entry was added after the upgrade!

One thing I don't really understand is that I couldn't find any entry on the logs about it being added 🤔 
(I've searched with `journalctl -g maintenance_window_start`) where should I look at @kyrofa ?

Pilzinsel64:
I wonder if there is anything else we missed. The script `1_add-maintenance-window-start.sh` just needs the execution bit. I wonder if git can commit that on linux (would be a surprise for me) or if there is another script we need to adjust (would also be a surprise as @kyrofa did good work in the past)?

Anisotropic:
Hi. There is "git update-index --chmod=+x path/to/file" command to store this "execution bit" in git. In details here: https://stackoverflow.com/questions/40978921/how-to-add-chmod-permissions-to-file-in-git

scubamuc:
fresh install tests completed with issues

grafik

grafik

ran usual tests as per [testing environmenmt](https://github.com/nextcloud-snap/nextcloud-snap/wiki/Develop-and-contribute#setup-testing-environment-per-script-optional) without setting the maintenance window as per [FAQ](https://github.com/nextcloud-snap/nextcloud-snap/wiki/FAQ's#q-how-can-i-set-default-maintenance-window-in-nextcloud-snap).

grafik

Pilzinsel64:
@scubamuc thank you for the test. But sadly it is needed to compile it locally like @pachulo explained. The reason that your test did not work is because the published snap is named like 27.0.4+gitxyz and will never trigger the new script.

src/migrations/migrations/28.0.0snap1/1_add-maintenance-window-start.sh
```diff
@@ -0,0 +1,22 @@
...
+# New config text
+insert_text="  'maintenance_window_start' => 1,"
+
+# Check if config name exists
+if grep -q "$search_text" "$config_file"; then
+    echo "The text '$search_text' is already available in the config file '$config_file'."
+else
+    # Paste the the default config value in the last line
+    sed -i '$i\'"$insert_text" "$config_file"
+    echo "The text '$insert_text' has been added to the config file '$config_file'."
```
pachulo, line 21:
```suggestion
    echo "The config '$insert_text' has been added to '$config_file'."
```

pachulo:
>  Also, given the risk of sedding config files, I'd like to see [shellspec tests](https://github.com/nextcloud-snap/nextcloud-snap/tree/master/tests/unit) for this.

@kyrofa I've been checking https://github.com/nextcloud-snap/nextcloud-snap/blob/master/tests/unit/common_utilities_spec.sh for inspiration on how the shellspec tests for this should look like, but I'm not really sure... Could you lend us a hand here?
Thanks!

src/migrations/migrations/28.0.0snap1/1_add-maintenance-window-start.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# Starting from Version 28 Nextcloud recommends to set a default time window for resource heavy background jobs.
+# See https://docs.nextcloud.com/server/28/admin_manual/configuration_server/background_jobs_configuration.html#maintenance-window-start
+
+# Path to config.php file
+config_file="/var/snap/nextcloud/current/nextcloud/config/config.php"
```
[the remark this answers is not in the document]
kyrofa: Yes, definitely. Although your suggested edit will introduce an extra quote.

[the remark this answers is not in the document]
pachulo: Thanks, updated!

Pilzinsel64:
Oh dear, it took the whole day to compile it, but it's done! Aaaand it finally works for me following the snaps from @pachulo (with one or two extra steps, I will make a wiki entry on how to compile somewhen the next few days).

The shell specs found a problem, whenever I try to fix it via this solution, I get new problems. :/

grafik
grafik

The Git history needs a cleanup. I will do when we fixed the issue found by the shell spec. :/

src/migrations/migrations/28.0.0snap1/1_add-maintenance-window-start.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# New config text
+insert_text="'maintenance_window_start' => 1,"
+
+# Check if config name exists
+if grep -q "$search_text" "$config_file"; then
+    echo "The config '$search_text' is already specified in '$config_file'."
+else
+    # Paste the the default config value in the last line
+    sed -i '$i\'"$insert_text" "$config_file"
```
pachulo, line 20: Can you explain the logic behind this command @Pilzinsel64 ? My bash foo is a little bit rusty 😓

kyrofa:
Hey guys, I'll be on vacation in a few days and should be able to pitch in a little here.

src/migrations/migrations/28.0.0snap1/1_add-maintenance-window-start.sh
Pilzinsel64: ...and mine non existent. Wrote that too lang ago. 😄
As far as I remember does `sed` repalce a text in-place (parameter `-i`) in the file `"$config_file"`. `$i` means to insert the new text (``) at before the last line. While `\` should somewhow escape the `"`. I'm not sure how to escape it else plus keeping the variable.

src/migrations/migrations/28.0.0snap1/1_add-maintenance-window-start.sh
```diff
@@ -0,0 +1,22 @@
...
+search_text="maintenance_window_start"
+
+# New config text
+insert_text="'maintenance_window_start' => 1,"
+
+# Check if config name exists
+if grep -q "$search_text" "$config_file"; then
+    echo "The config '$search_text' is already specified in '$config_file'."
+else
+    # Paste the the default config value in the last line
```
pachulo, line 19:
```suggestion
    # Paste the default config value in the last line
```

pachulo:
By the way, @kyrofa will this also add the config for people who install a fresh nextcloud >= 28? :thinking:

Pilzinsel64:
> will this also add the config for people who install a fresh nextcloud >= 28?

As far I know from my tests, this does not get executed on first fresh install. If you do a fresh install of NC28 and then do a refresh, it get executed.

kyrofa:
Hey @Pilzinsel64 I pushed the basic unit test scaffolding required to test this feature. I haven't fleshed out the tests yet, they will fail, but hopefully you can see where I'm going here. We need to make sure this is robust in a number of scenarios so we don't garble configs.

Pilzinsel64:
Ok, found a way now. Now I need to fix the tests, as it looks like. But I can't until Sunday, as I'm not at home until Sunday.

scubamuc (approved):

Pilzinsel64:
I'm done. Tests are happy and the script does what it should do. If you have any other suggestions I missed, just tell me.

tests/unit/migrations/add_maintenance_window_start_spec.sh
```diff
@@ -0,0 +1,50 @@
...
+	{
+		rm -rf "$SPEC_TMPDIR"
+	}
+
+	It 'handles empty config file'
+		When call "$MIGRATION"
+		The status should be failure
+	End
+
+	It 'handles basic config file'
```
kyrofa, line 28: I'd like to see three more test cases, and I think you'll see this still needs a smidge more work:

1. `It 'handles one line config file'`, using the following config:
```
?php
$CONFIG = array();
```

2. `It 'handles more complex config file'`, using the following config:
```
?php
$CONFIG = array(
'memcache.locking' => '\OC\Memcache\Redis',
'memcache.local' => '\OC\Memcache\Redis',
);
```

3. `It 'handles broken config file'`, using the following config (it should barf on this one instead of mangling it further):
```
?php
$CONFIG = array(
```

kyrofa: I think we might get stuck on this in that it almost requires a PHP interpreter written in bash. Maybe we should use nextcloud-fixer instead, where we can use OCC to do this.

Pilzinsel64: You mean [here](https://github.com/nextcloud-snap/nextcloud-snap/tree/master/src/nextcloud/fixes/existing-install) and [here](https://github.com/nextcloud-snap/nextcloud-snap/tree/master/src/nextcloud/fixes/fresh-install)?
On fresh install it just can add the config. On existing it would need to check if it already exists and only add if not exists.

kyrofa: Exactly. For the fresh install, though, we can probably just put it straight [in here](https://github.com/nextcloud-snap/nextcloud-snap/blob/master/src/nextcloud/config/config.php).

Pilzinsel64:
This script works on my private cloud.
grafik

Going to test three cases when the build is done and has been published:
- [ ] Does the config gets added on a fresh install?
- [ ] Does the config gets added on an upgrade from Nextcloud 27 without the config beeing available?
- [ ] Does the config doesn't get added on an upgrade from Nextcloud with the config beeing available?

tests/unit/migrations/add_maintenance_window_start_spec.sh
Pilzinsel64: Changed it to use migrations. I think it's fine as step 7 here without having maintenance mode enabled. I checked the used occ commands but there isn't a more easy way to add a config value only if the config value is empty.

But two more questions here:
1. What about tests? Didn't see tests for the migration steps (maybe I'm just blind).
2. Should I also revert the change from [tests/unit/spec_helper.sh](https://github.com/nextcloud-snap/nextcloud-snap/pull/2659/files#diff-93c02d4295748a9832e89fbc425d8fb0be7311652ca219a680970ba947adedcdL15) or do you want to keep them?

Pilzinsel64:
Ok, tested now and only a fresh install works. Seems like the script doesn't get executed on an upgrade.
Any ideas? Did i miss something? 🤔

pachulo:
> Any ideas? Did i miss something? 🤔

I think that the `src/nextcloud/fixes/existing-install/7_add_maintenance_window.sh` lacks [the execute bit](https://github.com/nextcloud-snap/nextcloud-snap/pull/2659#issuecomment-2060107808).

Pilzinsel64:
Oh comme on... Yea, that's it. Will fix it later. 🤦

Pilzinsel64:
All my tests were successful! 😃 Now I want to request review form one of you once again.

scubamuc:
fresh install tests completed 

`snap install nextcloud --channel=latest/beta/pr-2659`

grafik

grafik

- [ x ]  **default maintenance window was set correctly** 

grafik

tests/unit/migrations/add_maintenance_window_start_spec.sh
pachulo: > What about tests?

I would say they are not needed anymore, as we are using the official tool to update the config and not a bash script.

pachulo:
> All my tests were successful! 😃 Now I want to request review form one of you once again.

My opinion is that we can already merge this and then make a call for testing in https://forum.snapcraft.io/, what do think @kyrofa ?

Besides, to properly test this myself I need for it to be merged and then build into the `latest/beta` channel, because my production instance is arm64 based.

Thanks a lot for your tenacity @Pilzinsel64 !

pachulo:
gently ping @kyrofa

tests/unit/spec_helper.sh
```diff
@@ -12,7 +12,7 @@ spec_helper_precheck()
   # Available variables: VERSION, SHELL_TYPE, SHELL_VERSION
   : minimum_version "0.28.1"
   setenv SNAP="$SHELLSPEC_PROJECT_ROOT/src/common"
-  setenv NEXTCLOUD_SCRIPT_INCLUDE_PATH="src/common/utilities:src/apache/utilities:src/https/utilities:src/php/utilities:src/hooks/utilities"
+  setenv NEXTCLOUD_SCRIPT_INCLUDE_PATH="src/common/utilities:src/apache/utilities:src/https/utilities:src/php/utilities:src/hooks/utilities:src/nextcloud/utilities"
```
kyrofa, line 15: This is no longer necessary.

kyrofa:
Yes, this looks excellent, sorry for sending you on a wild goose chase here, but I think we've arrived at the proper solution.

>  Does the config gets added on a fresh install?
> Does the config gets added on an upgrade from Nextcloud 27 without the config beeing available?
> Does the config doesn't get added on an upgrade from Nextcloud with the config beeing available?

Just to confirm, one of these last two tested that we don't overwrite an existing `maintenance_window_start`? Take a quick glance at the logs of the nextcloud-fixer in these situations, too: does getting a key that doesn't exist look like an error in the log? If so, we might receive issues logged for that, and we might consider silencing such a thing. In all though, I'm +1 to this.

Pilzinsel64:
> Just to confirm, one of these last two tested that we don't overwrite an existing `maintenance_window_start`?

Yes, tested this too. I used config value `5` beeing on `latest/stable` (so NC27) and then refreshed to `latest/beta/pr-2659`.

> does getting a key that doesn't exist look like an error in the log

No, it doesn't. It just return an exit code `1` wich is catched by the if. But to confirm nothing is in the fixer-log, I quickyl need to install it once again later (need to go out now).

Pilzinsel64:
> > does getting a key that doesn't exist look like an error in the log
> 
> No, it doesn't. It just return an exit code `1` wich is catched by the if. And while testing the sh I didn't get an error. But to ensure nothing is in the fixer-log, I quickyl need to install it once again later (need to go out now).

Oh, you mean in the Nextcloud log? No, there isn't an error. Also, that would be increadible as the same function behind is used by Nextcloud itself covered in a function to get the default value if null.

kyrofa (approved):

kyrofa:
Awesome, let's ship it.

PR #2715, opened 2024-04-06
Release 27.1.8snap1

pachulo:


pachulo:
I've been using nextcloud 27.1.8 version on my instance for almost a week without a problem.
The apache HTTP upgrade was also flawless.
I think we should be releasing this, as 28.0.4 is still pending on how we sort out https://github.com/nextcloud-snap/nextcloud-snap/pull/2659

scubamuc (approved):

PR #2735, opened 2024-05-01
Use cmake snap to build MySQL 8.0.37

### LINKED ISSUE #2734: MySQL starting with 8.0.37 requires cmake 3.11.2 or newer
MySQL 8.0.37 requires a newer version of cmake than what comes with core18.  Should attempt to build cmake snap or pull in from source then reference this newer version to make MySQL.

stondino00:
Fixes #2734

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -250,6 +250,8 @@ parts:
       - libzip-dev
       - libargon2-0-dev
       - libsodium-dev
+      - libaio-dev
+      - libaio1
 
       # This is no longer bundled with PHP as of v7.4
       - libonig-dev
@@ -278,6 +280,7 @@ parts:
       - libargon2-0
       - libonig4
       - libsodium23
+      - libaio1
     prime:
      - -sbin/
      - -etc/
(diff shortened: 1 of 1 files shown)


stondino00:
@kyrofa let me know if you think I did this right pulling in the cmake snap or if it's still using the default cmake from core18.  Something like this might get us by until a rebase of a newer snap core.  Thanks sir.

Pilzinsel64 (approved):
I thinks that absolutely fine for now, so we can  continue pushing updates. We can partly revert this after we did the core20 update.

pachulo:
Great stuff @stondino00 ! One request though, could you clean the commit history?

scubamuc (approved):

stondino00:
> Great stuff @stondino00 ! One request though, could you clean the commit history? Thanks!

How do I do this via the web on github?  I don't like to use the github desktop or cli and do all the patching commits through the web.  I'm not much of a programmer.  I can close these and make a new clean PR if that would be easier.  Thanks

Pilzinsel64:
@stondino00 will clean it for you within the next two hours. It's the last thing I'll do before I am not at home until Sunday. ;)

PR #2739, opened 2024-05-09
ci: test only against protected branches

Pilzinsel64:
This will prevent duplicated tests if mistakenly created a PR from a nextcloud-snap branch. With this tests will only start on push to a protected branche `develop`, `main` and branches like `28` or `27`. It will also start tests on branches with an open Pull Request to one of the protected branches.

I'm using the same procedur for many of my mod repos: https://github.com/LITW-Refined/PilzMcSammelsorium/blob/master/.github/workflows/build-and-test.yml


kyrofa:
I'm not completely following why you want to do this. Are you wanting to avoid CI when you push to a development branch? Maybe just use a fork?

Pilzinsel64:
> Are you wanting to avoid CI when you push to a development branch?

Yes.

> Maybe just use a fork?

Basically yes. You even want to keep the branche lust clean, so mostly useless. But still see it sometimes. It's just a suggestion that took me less then two minutes. Feel free to close or discuss or what you feel about. :)

PR #2795, opened 2024-06-21
php: enable sqlite3 & pdo-sqlite

Pilzinsel64:
Fix #2744

Need to be tested, not sure if we need to enable something explicitely or if its enough to remove the disable lines.

PR #2839, opened 2024-07-23
Release 29.0.4snap2

pachulo:


pachulo:
> @pachulo is it an option to merge https://github.com/nextcloud-snap/nextcloud-snap/pull/2795 to 29.0.4snap2?

Not really. We first have to decide if we want to include the extra dependency so that one feature from one app works as expected.

HyknuF:
@pachulo 

Thanks for all your work and effort. 

At the moment i am struggling with the same issue. It seems there is no alternative app to collectives for integrated knowledgebase, so i am "forced" to use that. Yesterday i saw the "closed as completed issues" and thought: Yess! :-) But then i realised it isn't merged yet unfortunately. Having a search feature would be really nice, especially in large knowledgebases. I absolutely get your point not be keen to do that. Hopefully you can merge that into stable anyway. I would really appriciate that step
Kind regards!

PR #2859, opened 2024-08-23
Upgrade Redis to 7.2.6

### LINKED ISSUE #2855: Upgrade Redis to 7.2.5
Seems like our 7.0.X version has gone end of life.

image

stondino00:
Fixes #2855

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -294,9 +294,9 @@ parts:
 
   redis:
     plugin: redis
-    source: https://download.redis.io/releases/redis-7.0.15.tar.gz
-    source-checksum: sha256/98066f5363504b26c34dd20fbcc3c957990d764cdf42576c836fc021073f4341
-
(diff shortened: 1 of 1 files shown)


stondino00:
Builds fine but when I updated to this PR, nextcloud was not accessible.  Seems will need more investigation.

pachulo:
I've just tested the PR, and I see this error on the Redis log (/var/snap/nextcloud/current/logs/redis.log):

```
8683:M 27 Aug 2024 22:04:33.455 # Failed to configure LOCALE for invalid locale name.
```

Which makes me think that we will need to update the snap redis config using [this](https://raw.githubusercontent.com/redis/redis/7.2/redis.conf), because there is a new option in there related to the locale:

```
# Set the local environment which is used for string comparison operations, and 
# also affect the performance of Lua scripts. Empty String indicates the locale 
# is derived from the environment variables.
locale-collate ""
```

I will try to do it next week if nobody can do it before.

stondino00:
Closing as 7.2.6 has been released.

PR #2914, opened 2024-10-11
Enable Dependabot version update for actions

pachulo:
I think this is worth it, what do you think @kyrofa ?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+# Set update schedule for GitHub Actions
+
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


pachulo:
Can I go on with this @kyrofa ?

kyrofa:
Your call @pachulo, I have no informed opinion on this.

pachulo:
Well, I missed this one: this is opening the PRs against our default branch, `master`, but I'm of the opinion that we should [configure it](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#target-branch) to open them on `develop`. Do you agree @kyrofa ?

kyrofa:
Yes, agreed.

PR #3008, opened 2024-12-28
Make snap-name dynamic with aliasing support

jmgiaever:
Use dynamic snap naming to support aliasing

### DIFF
--- a/src/hooks/bin/configure
+++ b/src/hooks/bin/configure
@@ -68,7 +68,7 @@ handle_apache_port_config()
 	apache_set_previous_https_port "$https_port"
 
 	# Restart Apache to apply new config
-	snapctl restart nextcloud.apache
+	snapctl restart "${SNAP_INSTANCE_NAME}.apache"
 }
 
 handle_php_memory_limit()
@@ -93,7 +93,7 @@ handle_php_memory_limit()
 	php_set_previous_memory_limit "$memory_limit"
 
 	# Restart PHP to apply new config
-	snapctl restart nextcloud.php-fpm
+	snapctl restart "${SNAP_INSTANCE_NAME}.php-fpm"
 }
 
 handle_cronjob_interval()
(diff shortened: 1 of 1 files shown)


pachulo:
Hi @jmgiaever,

First, thanks for your contribution to the snap!

> support aliasing

Second, could you elaborate which use cases is this enabling?

jmgiaever:
Hi,

This way the snap can be installed and operate in parallel, e.g `sudo snap install nextcloud_second`.

pachulo (approved):

pachulo:
I think we need another approval for the checks to run. Am I right @kyrofa ?

kyrofa:
> I think we need another approval for the checks to run. Am I right @kyrofa ?

No, we just need to approve new contributors before pipelines runs ("approve" is overloaded, here).

kyrofa (changes requested):
Hey there @jmgiaever, thanks for the PR! I'm afraid I must say that I don't love it, but not because there's anything wrong with your work: it's because I absolutely detest how the parallel install feature exists in snapd: it foists work onto individual snaps by not inferring what it's talking about when it uses `<snap name>.<app name>`. We should not be needing to use `$SNAP_INSTANCE_NAME` at all here, and the fact that we do or we'll break parallel installs means that we're almost certain to accidentally break parallel installs in the future. Rant aside, I'm not opposed to this, just know that we don't utilize parallel installs and don't test them, so this may regress in the future.

Note that various commands use `nextcloud.<command>` in their usage text, please update those as well. `grep -r COMMAND=\"nextcloud\. src/` to find them.

Can you also please update your commit? This doesn't relate to aliasing, it relates to parallel installs.

pachulo:
Do you plan on working on this @jmgiaever ?

jmgiaever:
Hi,

Yes, but at the moment I'm a bit occupied with work, so I don't have time to shift focus. 

It's on my list, and I'll get back to you on it when my situation calms down in April. Hope that's ok!

pachulo:
Sure, not problem at all! We are all volunteers here ;)

Issue #2966, opened 2024-11-12
"Release snap for pull request" action broken


sbe-arg:
We need a better process for testing github actions so we don't break releases.

Ie. Don't merge actions upgrades along nextcloud upgrades

pachulo:
I don't think anything broke with the latest upgrade of an action, but we will see in the following PRs we open.

pachulo:
It seems like everything went fine over here: https://github.com/nextcloud-snap/nextcloud-snap/pull/2973

sbe-arg:
What is very confusing is the failures each time it runs the launchdpad action.

And because we can't publicly see the runner logs I cannot help trobleshoot.

Actions should be always be green if they are doing what is supposed to do. But unsure why we have so much release reds.

When we released 30.0.2 was red for the 30.0.2 tag run and there was no new version available in snap until a few days after.

Whats the release process did you re run something manually after? Or worker but ended in red fail anyway.

pachulo:
> What is very confusing is the are failures each time it runs the launchdpad action.

The [deploy job](https://github.com/nextcloud-snap/nextcloud-snap/blob/master/.github/workflows/launchpad.yml#L14) from the [Deploy to launchpad action](https://github.com/nextcloud-snap/nextcloud-snap/blob/master/.github/workflows/launchpad.yml) has a known problem, because we can't automatically push git tags to Launchpad repos (am I right @kyrofa ?).
So every time there's a new tag to be pushed, it fails:
<img width="1102" alt="image" src="https://github.com/user-attachments/assets/7bdd6c5c-ea7f-488e-bb40-e567451a398c">

So we have to push them manually.

> Actions should be always be green if they are doing what is supposed to do. But unsure why we have so much release reds.

I would say is a mix of things, but is true that a lot of [Release snap for pull request actions](https://github.com/nextcloud-snap/nextcloud-snap/blob/master/.github/workflows/release-pr-snap.yml) runs are failing lately:
<img width="1102" alt="image" src="https://github.com/user-attachments/assets/9ce8c047-e40f-4d1b-bd26-1b1f65b58edd">

It's because the [Upload snap step](https://github.com/nextcloud-snap/nextcloud-snap/blob/master/.github/workflows/release-pr-snap.yml#L111) never finishes:
<img width="1102" alt="image" src="https://github.com/user-attachments/assets/b46458e4-8d05-4dec-9511-3460c0123ecf">

Do you have any idea why @kyrofa ?

[...]

PR #3020, opened 2025-01-12
actions: workaround checkout action not fetching annotated tags

pachulo:
Second try at solving the problem we have with Launchpad just getting lightweight tags from the action. More context https://github.com/nextcloud-snap/nextcloud-snap/issues/2966#issuecomment-2502193356.


kyrofa (approved):
How annoying. Thanks @pachulo.

Issue #2908, opened 2024-10-09
Alternatives to snapcraft local plugins


pachulo:
I will try to document everything I know about the issue and how we could try to advance.

Could you fact-check it @kyrofa ? Thanks!

### Next steps / To-Do
- [x] Document everything we know about the local plugins in an issue.
- [ ] Check if I got everything right and/or missed something.
- [ ] Ask for help in [the snapcraft forum](https://forum.snapcraft.io/).
- [ ] Ask for help in [one of the starcraft clinics that the snapcraft team organizes](https://docs.google.com/document/d/1KHCaIU5VxA6UKSW5hD0zxFqUcZNvTc7_8fDRJ2Ihpu8/edit).

### Current situation
For building some of the nextcloud snap parts, we are using a snapcraft feature called (v1) [local plugins](https://forum.snapcraft.io/t/writing-local-plugins/5125).

The source-code of the three local plugins is located [here](https://github.com/nextcloud-snap/nextcloud-snap/tree/master/snap/plugins) and the rationale for using local plugins, instead of the official ones, is explained in the following blog posts by @kyrofa :
* [Apache](https://kyrofa.com/posts/snapping-nextcloud-the-web-server/)
* [PHP](https://kyrofa.com/posts/snapping-nextcloud-php-7/)
* [Redis](https://kyrofa.com/posts/snapping-nextcloud-the-memcache/)

### Problem
Because the use of local plugins, the nextcloud snap is currently stuck on `core18`, and that is already [causing](https://github.com/nextcloud-snap/nextcloud-snap/issues/2335#issue-1602769071) [problems](https://github.com/nextcloud-snap/nextcloud-snap/issues/2808).
Besides, even if we upgraded to v2 local plugins [they are only supported on core20-based snaps, as the whole concept is deprecated on core22 onwards](https://forum.snapcraft.io/t/local-v2-plugin-base-core22/32227).
To make the situation worse, there are [no plans to re-introduce something similar in the foreseeable future](https://forum.snapcraft.io/t/local-v2-plugin-base-core22/32227/13).

So, if we don't find any alternative, the nextcloud snap will be, at best, stuck in `core20` indefinitely.

### Possible solutions
The alternatives we have were explained [here](https://forum.snapcraft.io/t/local-v2-plugin-base-core22/32227/13) by @mr-cal:
> 1. to use the nil plugin with override- keywords,
[...]

PR #3073, opened 2025-03-06
Upgrade to core22 (original contribution by soumya dghosh)

pachulo:
This is a first attempt at solving #2908 .

It is what @soumyaDghosh shared on https://github.com/nextcloud-snap/nextcloud-snap/pull/2542#issuecomment-2676738999.

We now need to make it compile and then verify if we get a functional snap.

If we do so, we will clean the diff.

~~For the time being, the build will be happening [on launchpad](https://launchpad.net/~pachulo/+snap/nextcloud-core22).~~
I've updated the PR so that Github CI/CD can also build this.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -302,18 +302,7 @@ parts:
       - --prefix=/usr
     override-build: |
       $CRAFT_STAGE/usr/bin/phpize
-
-      # Run the default build (configure and make)
-      # We skip 'craftctl default' to avoid the implicit 'make install' writing to the wrong place
-      ./configure --prefix=/usr
-      make -j$CRAFT_PARALLEL_BUILD_COUNT
-
-      # Calculate the PHP API version directory (e.g., no-debug-non-zts-20230831)
-      API_DIR=$($CRAFT_STAGE/usr/bin/php-config --extension-dir | xargs basename)
-
-      # Manually install the extension to the correct location in this part
-      # We override 'extension_dir' to force it into the standard /usr path structure
-      make install DESTDIR=$CRAFT_PART_INSTALL extension_dir=/usr/lib/php/extensions/$API_DIR
+      craftctl default
 
   redis:
     plugin: make


pachulo:
I guess we will need a new way of executing the integration tests for supporting the upgrade to core22.

pachulo:
For the time being, the build will be happening [on launchpad](https://launchpad.net/~pachulo/+snap/nextcloud-core22).

sbe-arg:
Where are we at with this pr?

sbe-arg:
If we stay on core18 we have security updates till about april-may 2028 right?

Just triple checking that part.

sbe-arg:
bump

snap/snapcraft.yaml
```diff
@@ -277,25 +285,33 @@ parts:
...
+      - -etc/
     organize:
       # Put php-fpm in bin with everything else, not sbin.
-      sbin/php-fpm: bin/php-fpm
-    extensions:
-      # Build the redis PHP module
-      - source: https://github.com/phpredis/phpredis/archive/6.1.0.tar.gz
-        source-checksum: sha256/57135db32a0ccb1659f56c75feb26c10ea94fb3d2471edd047d94a9800f959b0
+      usr/sbin/php-fpm: usr/bin/php-fpm
+      usr/lib/php/extensions/no-debug-non-zts-20220829: usr/lib/php/extensions/
```
joshtrichards: Path needs to be update for the PHP API version bump:

```suggestion
      usr/lib/php/extensions/no-debug-non-zts-20230831: usr/lib/php/extensions/
```

snap/snapcraft.yaml
```diff
@@ -277,25 +285,33 @@ parts:
...
+  redis-php:
+    after: [php]
+    source: https://github.com/phpredis/phpredis/archive/6.1.0.tar.gz
+    source-checksum: sha256/57135db32a0ccb1659f56c75feb26c10ea94fb3d2471edd047d94a9800f959b0
+    plugin: autotools
+    autotools-configure-parameters:
+      - --prefix=/usr
+    override-build: |
+      $CRAFT_STAGE/usr/bin/phpize
+      craftctl default
```
joshtrichards, line 305: This *may* fix the build[^1]. I think it's at least part of the solution (to make sure php-redis, which is built independently from php and the other built-in php extensions).

```suggestion
      craftctl default
   organize:
      usr/lib/php/extensions/no-debug-non-zts-20230831: usr/lib/php/extensions/
```

[^1]: From build logs (https://launchpadlibrarian.net/781440589/buildlog_snap_ubuntu_jammy_amd64_nextcloud-core22_BUILDING.txt.gz): `Failed to copy '/build/nextcloud/stage/usr/lib/php/extensions/no-debug-non-zts-20230831/redis.so': no such file or directory.`

snap/snapcraft.yaml
joshtrichards: Actually I take that back. This may need a different type of change, because `extension_dir` in the `php.ini` is versioned (i.e. has the API string in it). Hmmm... may be the cause of the build failure for redis-php actually: I'd just drop this line from the `php` part. And, in that case, there should be no need to add the `organize` / equivalent to `redis-php`. All extensions should end up landing in the correct (versioned) directly in that case (including redis-php).
```suggestion
```

pachulo: So, do you think this change is also needed @joshtrichards ?

joshtrichards: I'd say go without `usr/lib/php/extensions/no-debug-non-zts-[version]: usr/lib/php/extensions/` lines for both the `php` and `php-redis` parts. Dropping it from the former should remove the need for it to be added to the latter.  And also then the date version string doesn't have to be fixed either.

pachulo: OK, let's try it!

Pilzinsel64:
On `sudo snapcraft --provider lxd` in `test.yml`:
> Option --provider is not supported.

Use `--use-lxd` instead. This should solve at least the current issue.

```diff
--- .github/workflows/test.yml
+++ .github/workflows/test.yml
@@ -54,9 +54,9 @@
         run: sudo iptables -I DOCKER-USER -i lxdbr0 -o eth0 -j ACCEPT && sudo iptables -I DOCKER-USER -o lxdbr0 -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
 
       # Using sudo here because our CI user isn't a member of the lxd group
       - name: Build snap
-        run: sudo snapcraft --provider lxd
+        run: sudo snapcraft --use-lxd
 
       - name: Review snap
         run: sudo snap install review-tools && /snap/bin/review-tools.snap-review *.snap
```

pachulo:
@joshtrichards this is what I see when I debug this on my local machine:
```
$ snapcraft pack --use-lxd --debug
Failed to copy '/root/stage/usr/lib/php/extensions/no-debug-non-zts-20230831/redis.so': no such file or directory.
Launching shell on build environment...
snapcraft-nextcloud-on-arm64-for-arm64-262314 ../project# find / -iname redis.so
/root/parts/php/install/usr/lib/php/extensions/no-debug-non-zts-20230831/redis.so
/root/parts/redis-php/build/.libs/redis.so
/root/parts/redis-php/build/modules/redis.so
```

Any ideas?

pachulo:
We will try to solve this step by step instead, see [here](https://github.com/nextcloud-snap/nextcloud-snap/pull/3373#issuecomment-3889724687).

PR #3092, opened 2025-03-18
run ci only on protected branches

Pilzinsel64:
Does what it says. Made a PR a year ago already but it didn't get merged or reviewed. Another try, now that I know @kyrofa would accept it probably. :)

It's the same way like I've seen for ohter organizations.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,16 @@
 name: Run tests against changes
 
-on: [push, pull_request]
+on:
+  pull_request:
+    branches:
+      - '*'
+  push:
+    branches:
+      - master
+      - develop
+      - '[0-9]+'
+    tags:
+      - '*'
 
 jobs:
   static:


scubamuc (approved):

.github/workflows/test.yml
```diff
@@ -1,6 +1,16 @@
...
 
-on: [push, pull_request]
+on:
+  pull_request:
+    branches:
+      - master
+      - develop
+      - '[0-9]+'
+  push:
+    branches:
```
kyrofa, line 10: This looks fine. Should we add a `tags: [*]` as well?

Pilzinsel64: If we want to run CI on tag pushes aswell, then yes, of course!

.github/workflows/test.yml
```diff
@@ -1,6 +1,16 @@
 name: Run tests against changes
 
-on: [push, pull_request]
+on:
+  pull_request:
+    branches:
+      - master
+      - develop
+      - '[0-9]+'
```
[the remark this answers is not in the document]
Pilzinsel64: Yea, true, here we should also just allow all branches. You would be able to accept PRs to WIP branches then. Totally fine. And if you want to run CI on your banche, just make a (draft) PR to `develop`. This way I've seen in several orgs already.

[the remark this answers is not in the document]
kyrofa: Yeah, let's do that.

.github/workflows/test.yml
kyrofa: Perfect.

kyrofa (approved):

kyrofa:
Nothing has changed at all. We still use a forking feature-based workflow. But at least now if someone doesn't use their own fork and pushes directly to the main project, we don't get duplicate CI jobs.

PR #3305, opened 2025-10-28
Release 32.0.1snap1

pachulo:


Pilzinsel64 (approved):
Works fine in prod for several days now. No issues so far.

scubamuc:
@Pilzinsel64 it won't take but a day before the first issue is created complaining that the install failed using sqlite or postgress or anything other than mysql... i'm quite sure that this has been an issue way back when... @kyrofa can you remember?

scubamuc:
> If they actually agreed that this was a regression to be fixed, I'd say let's hold off. But it looks like they don't necessarily agree that it's a regression, is that right? It may stay this way.

yes, that seems to be the case as mentioned by @kesselb 
if so it would require editing documentation and screen shots upstream in official install manual, forum documentation, wiki documentation and readme.md... not sure how to stem that 😵

scubamuc:
@nextcloud-snap/developers we need confirmation https://github.com/nextcloud/server/issues/56160#issuecomment-3495505923 for the upstream folks @skjnldsv to go ahead

PR #3344, opened 2025-12-25
php: enable phar

pachulo:
Let's see if this is the only dependency that we need in PHP for the ExApps to work.


sbe-arg (approved):

pachulo:
> testing approval and confirming https://github.com/nextcloud-snap/nextcloud-snap/issues/3211#issuecomment-3712859149

OK, I guess we can then merge this, as it is a hard requirement to run ExApps, while we try to find out how to solve the "Redis not available for local cache" error.

PR #3404, opened 2026-03-18
Upgrade to core20

christianlupus:
This is a PR just to update to the minimal compatible versions of the different parts.

Let's see if it builds and can be installed after that.

PR #3403, opened 2026-03-17
Release 32.0.6snap3

pachulo:

Issue #1012, opened 2019-05-20
cerbot: re-introduce webroot plugin


kyrofa:
#1008 (rightly) removed the custom webroot plugin (it's no longer needed these days). However, this breaks existing cert renewals since that config option is saved. Need to revert that move for now. Continue using certbot's plugin for obtaining certs, though.

PR #3405, opened 2026-03-19
Upgrade to core20 + drop Certbot manual config

christianlupus:
This is just an idea on one could avoid building the certbot manually and instead use the preexisting certbot package

This is based on #3404.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -136,6 +136,7 @@ parts:
       - libaprutil1-dev
       - libpcre2-dev
       - libssl-dev
+      - rsync
 
     stage-packages:
       - libbrotli1
@@ -144,6 +145,7 @@ parts:
       - libpcre2-8-0
     
     autotools-configure-parameters:
+      - --prefix=/
       - --with-mpm=event
       - --enable-modules=none
       - --enable-mods-static="headers proxy proxy_fcgi setenvif env rewrite mime dir authz_core unixd alias ssl socache_shmcb slotmem_shm log_config logio brotli filter"
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -15,6 +15,9 @@ architectures:
   - build-on: armhf
   - build-on: ppc64el
 
+environment:
+  PYTHONPATH: $SNAP/usr/lib/python3.*:$SNAP/usr/lib/python3.*/lib-dynload:$SNAP/usr/lib/python3/dist-packages
+
 apps:
   # Apache daemon
   apache:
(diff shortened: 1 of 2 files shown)


snap/snapcraft.yaml
```diff
@@ -430,16 +432,13 @@ parts:
...
+    plugin: nil
+    stage-packages: [certbot]
     after: [patches]
-    override-build: |
-      snapcraftctl build
-      patch -p1 -d $SNAPCRAFT_PART_INSTALL/lib/python2.7/site-packages/certbot < $SNAPCRAFT_STAGE/certbot-remove-default-config-files.patch
-      patch -p1 -d $SNAPCRAFT_PART_INSTALL/lib/python2.7/site-packages/certbot < $SNAPCRAFT_STAGE/certbot-remove-storage-chown.patch
+    #override-build: |
+    #  snapcraftctl build
+    #  patch -p1 -d $SNAPCRAFT_PART_INSTALL/lib/python3*/site-packages/certbot < $SNAPCRAFT_STAGE/certbot-remove-default-config-files.patch
```
christianlupus, line 440: These patches can be incorporated as well. I am not sure if we really need them. A quick test showed no need for them. Thus, I kept them out. Can be adjusted.

christianlupus:
One issue, that might arise is #1012: I do not know what happened to the old config using the custom nextcloud webroot plugin. Just to keep that in mind

scubamuc:
fresh install tests completed without issues

nextcloud (beta/pr-3405) 0+git.65a54fd from Nextcloud✓ installed

image

running snapd core20
image

image

image

image

certbot help file
image

certbot get certificate
image

certbot logs

image

scubamuc (approved):

scubamuc:
@christianlupus 

> The problem is: If a user installed a snap at the time this plugin was necesary and used, the information that the cert uses the "nextcloud plugin" is stotred in the renew config. Thus, every renew will use that same plugin as well if I am not mislead. This, Kyle dropped the plugin and reinserted it just the day after. So, these installations might be out there running still on the ancient plugin.

@kyrofa any suggestions? there shouldn't be any snaps using that ancient plugin since [32.0.6snap3](https://github.com/nextcloud-snap/nextcloud-snap/pull/3403) has 0.39 certbot already?

kyrofa:
> there shouldn't be any snaps using that ancient plugin

There definitely were back then. Unless all of our older users have left, they're probably still using it. Those are the users who really love us, so I would be loathe to break them at this point. I wonder if we can make a [migration](https://github.com/nextcloud-snap/nextcloud-snap/tree/master/src/migrations) and sed the config file to use the newer plugin. That would require some extensive testing, though-- someone would need to get a cert with the old plugin so we know exactly what it looks like and then test the migration script using it, ensuring they can renew the cert with the new plugin afterward.

christianlupus:
> I wonder if we can make a [migration](https://github.com/nextcloud-snap/nextcloud-snap/tree/master/src/migrations) and sed the config file to use the newer plugin. That would require some extensive testing, though-- someone would need to get a cert with the old plugin so we know exactly what it looks like and then test the migration script using it, ensuring they can renew the cert with the new plugin afterward.

Exactly, what I had in mind: using a migration to get these users to the new plugin. The question would be: is it easier to patch the config file and let certbot hopefully handle the rest or plainly drop the old certificate (as part of the update) and issue a new one automatically. That way, we would only check if the plugin is in use and not tweak the file in potentially dangerous ways. Certbot knows best what to store in the config files.

kyrofa:
> [...] or plainly drop the old certificate (as part of the update) and issue a new one automatically

That requires us to track the info they used when they originally called certbot (domains, email addresses, etc.), which we haven't needed to do. All of that state is contained in those config files, so I think sed is our only real option. Your point is still valid though, this is totally dangerous.

christianlupus:
I thought to extract the data from the config file (yeah sed might be an option) but then let certbot modify the actual files.

kyrofa:
Fair, that's doable. Although then it's possible that we run into rate limit issues trying to force a new cert. I wonder if there's just a way to get certbot to generate the config and not actually do anything else.

christianlupus:
I honestly doubt there is an option for config only, to be honest. But I did not check. 

I am thinking a different direction: unless a hacky user used `snap run` to create new certificates, there cannot be more than 1 renews in the past week. You have 5 requests or so within 1 week with the exact same set of domains. So, there should be plenty of reserves. 
Yeah, a power user might run into a ip-based rate limit. But these are way higher and also reset rather quickly. If renewal does not work successfully, we could just restore the certificate...

pachulo:
@christianlupus can we close this, now that we have https://github.com/nextcloud-snap/nextcloud-snap/issues/3419?

christianlupus:
I'd say yes. I will keep the branch around just in case.

PR #3406, opened 2026-03-20
Allow certbot force renewal

christianlupus:


christianlupus:
I formally marked this as ready for review. Let's have a discussion on what is needed to get this done. I see this as a preparation to the upgrade agenda we do have (to core24 or 26) in the upcoming time. So, let's see what your opinions are.

christianlupus:
Since there was not much feedback on this PR and the use case is more related to testing and less about actual usage, I am no longer convinced, we need this. Should we simply close and forget it?

PR #3440, opened 2026-04-11
actions: use snapcraft 8 for building the snap

pachulo:


Pilzinsel64 (approved):
Changes seems legit to me, what I expect from what I know in the past when manually building it locally.

scubamuc (approved):

PR #3445, opened 2026-04-13
mysql: enable mysql_native_password

pachulo:
This will be deprecated in MySQL 9.0, but it's still available in version 8.4.
We should do [the proper migration though](https://php.watch/articles/fix-php-mysql-84-mysql_native_password-not-loaded).

### DIFF
--- a/src/mysql/my.cnf
+++ b/src/mysql/my.cnf
@@ -7,3 +7,4 @@ skip-log-bin
 transaction_isolation=READ-COMMITTED
 log_error=../logs/mysql_errors.log
 performance_schema=OFF
+mysql_native_password=ON


christianlupus:
It is not only deprecated but removed. So, we must handle the actual migration during 8.4.

pachulo:
> It is not only deprecated but removed. So, we must handle the actual migration during 8.4.

Look at the [documentation](https://dev.mysql.com/doc/refman/8.4/en/native-pluggable-authentication.html):
> The mysql_native_password authentication plugin is deprecated as of MySQL 8.0.34, disabled by default in MySQL 8.4, and removed as of MySQL 9.0.0.

And the doc was right: I was able to update my production instance to MySQL 8.4 with this build! 🥳 

@stondino00 can you try the upgrade with this PR, to see if it also solves the issue for you?

And, taking into account that everything goes fine with @stondino00 instance: what do you think about the following plan @nextcloud-snap/developers & @christianlupus ?
1. Release 32.0.8snap3 with the upgrade to MySQL 8.4 and this config.
2. Create an issue for the migration of the DB user to caching_sha2_password, which we will need to implement before the upgrade to MySQL 9.0.

kyrofa:
@pachulo is correct, but I suggest doing the migration now, while we're thinking about it, rather than blocking future work on it. Given my brief research, we'll probably need this config tweak anyway because I think we need to be able to run a mysql shell command in order to do this migration, but then we can conceivably drop it (maybe it's something we can specify on the CLI instead, that'd be handy).

christianlupus:
I agree with @kyrofa, it makes sense to

1. Use this config to allow access to MySQL temporaryly
2. Migrate to caching_sha2_password 
3. Get rid of the config as it will break with MySQL 9.

This should be done in one (migration) step before the currently broken 8.4 is further developed. Just my 50ct

pachulo:
> Given my brief research, we'll probably need this config tweak anyway because I think we need to be able to run a mysql shell command in order to do this migration, but then we can conceivably drop it 

Indeed: I think we will need a way to connect to the DB service, whatever we do.

> )maybe it's something we can specify on the CLI instead, that'd be handy).

I also don't know if that's possible.

So, @kyrofa & @christianlupus do you both agree on releasing 32.0.8snap3 with both the MySQL 8.4 update and this config, and then start working on the caching_sha2_password  migration?

I just want to highlight that right now I don't have a lot of free time to work on that migration, but I can help with testing.

kyrofa:
> So, @kyrofa & @christianlupus do you both agree on releasing 32.0.8snap3 with both the MySQL 8.4 update and this config, and then start working on the caching_sha2_password migration?

Sorry for being unclear. I recommend _against_ this approach because our immediate incentive for doing this disappears, which means we will in all likelihood forget about it until we need to ship v9 and then hit a brick wall, needing to come fully up to speed on the problem again. We'll also forget all about this config option that should be removed. I don't mean to criticize us, this is just how things work sometimes! This also completely ignores the fact that MySQL made this change for a reason, likely security-related, and it's probably in our users best interests to make this happen. That's not to put all the pressure on you though @pachulo, let's see if someone else can potentially pick up the migration.

Pilzinsel64:
In this case I agree with all of you. But for the next steps, I'm on board with @pachulo and @christianlupus. Maybe let's do a small roadmap, as long as we follow it, we wont miss it.

So, basically this:
1. Keep the mysql update with the option in #3445 -> Release
2. Update Nextcloud -> Release
3. Make a migration & remove the option -> Release
4. +5. Upgrade MySQL / core20 to core22

While 1. and 2. already think about or draft a PR for the migration.

> I suggest doing the migration now

Yes, but step-wise. What I said above. Like this we can update MySQL, keep it secure. Also keep Nextcloud (and potentially other dependency updates) updated and secure. And also like this we wont miss the migration.

pachulo:
Closing in favor of https://github.com/nextcloud-snap/nextcloud-snap/pull/3446

PR #3446, opened 2026-04-13
mysql: migrate users to caching_sha2_password plugin

kyrofa:
The `mysql_native_password` plugin has been deprecated for a while, but was finally disabled by default in MySQL v8.4, which made it catch our attention. Some long-running installations will still be using `mysql_native_password`; migrate those installations to `caching_sha2_password`.

`mysql_native_password=ON` needs to be in the config in order to actually complete this migration. We can remove it in a future update.

### DIFF
--- a/src/nextcloud/fixes/existing-install/maintenance/3_mysql-use-caching_sha2_password.sh
+++ b/src/nextcloud/fixes/existing-install/maintenance/3_mysql-use-caching_sha2_password.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+#
+# Version 32.0.8snap3 introduced MySQL v8.4, which disables `mysql_native_password`
+# by default. Let's make sure that our users are migrated to use
+# `caching_sha2_password`.
+set -e
+
+# shellcheck source=src/mysql/utilities/mysql-utilities
+. "$SNAP/utilities/mysql-utilities"
+
+root_password="$(sed -rn 's/password=(.*)/\1/p' "$MYSQL_ROOT_OPTION_FILE")"
+nextcloud_password="$(mysql_get_nextcloud_password)"
+
+"$SNAP"/bin/run-mysql <<SQL
+ALTER USER 'root'@'localhost' IDENTIFIED WITH caching_sha2_password BY '$root_password';
+ALTER USER 'nextcloud'@'localhost' IDENTIFIED WITH caching_sha2_password BY '$nextcloud_password';
+SQL
--- a/src/mysql/my.cnf
+++ b/src/mysql/my.cnf
@@ -7,3 +7,4 @@ skip-log-bin
 transaction_isolation=READ-COMMITTED
 log_error=../logs/mysql_errors.log
 performance_schema=OFF
+mysql_native_password=ON


kyrofa:
I haven't actually tested this myself, I'm afraid, I just wanted to get it out there for us to iterate on. @pachulo would you mind taking this for a spin once it builds? I'm assuming I can ignore the `mysql.session` and `mysql.sys` users... :crossed_fingers:

src/nextcloud/fixes/existing-install/maintenance/3_mysql-use-caching_sha2_password.sh
```diff
@@ -0,0 +1,17 @@
...
+# Version 32.0.8snap3 introduced MySQL v8.4, which disables `mysql_native_password`
+# by default. Let's make sure that our users are migrated to use
+# `caching_sha2_password`.
+set -e
+
+# shellcheck source=src/mysql/utilities/mysql-utilities
+. "$SNAP/utilities/mysql-utilities"
+
+root_password="$(sed -rn 's/password=(.*)/\1/p' "$MYSQL_ROOT_OPTION_FILE")"
+nextcloud_password="$(mysql_get_nextcloud_password)"
```
kyrofa, line 12: Given that this is a change in the default hashing strategy, it appears that we need to provide the plaintext password. We have the one we used to generate the installation, but this assumes that the user hasn't changed it out from under us. Seems like a fairly safe assumption given that we embed MySQL within the snap, but one I wanted to call out anyway.

pachulo:
Morning! I've already tried the update and it seems to work great!

Here is the execution of the migration:

```
...
Apr 14 05:51:25 nextcloud-container nextcloud.nextcloud-fixer[324481]: run-parts: executing /snap/nextcloud/53052/fixes/existing-install/maintenance/3_mysql-use-caching_sha2_password.sh
Apr 14 05:51:25 nextcloud-container nextcloud.nextcloud-fixer[320897]: Disabling maintenance mode...
Apr 14 05:51:26 nextcloud-container nextcloud.nextcloud-fixer[320897]: done
Apr 14 05:51:29 nextcloud-container systemd[1]: snap.nextcloud.nextcloud-fixer.service: Deactivated successfully.
Apr 14 05:51:29 nextcloud-container systemd[1]: snap.nextcloud.nextcloud-fixer.service: Consumed 34.139s CPU time, 11.2M memory peak, 0B memory swap peak.
```

And here the table:

```
ubuntu@nextcloud-container:~$ sudo nextcloud.mysql-client
Welcome to the MySQL monitor.  Commands end with ; or \g.
Your MySQL connection id is 47
Server version: 8.4.8 Source distribution

Copyright (c) 2000, 2026, Oracle and/or its affiliates.

Oracle is a registered trademark of Oracle Corporation and/or its
affiliates. Other names may be trademarks of their respective
owners.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

mysql> SELECT user, host, plugin from mysql.user WHERE plugin='mysql_native_password';
Empty set (0.00 sec)

mysql> SELECT user, host, plugin from mysql.user;
+------------------+-----------+-----------------------+
| user             | host      | plugin                |
+------------------+-----------+-----------------------+
| mysql.infoschema | localhost | caching_sha2_password |
| mysql.session    | localhost | caching_sha2_password |
| mysql.sys        | localhost | caching_sha2_password |
| nextcloud        | localhost | caching_sha2_password |
| root             | localhost | caching_sha2_password |
+------------------+-----------+-----------------------+
5 rows in set (0.00 sec)
```

So this looks quite solid to me.

@stondino00 would you be able to test this when you have a chance?

Thanks!

christianlupus:
Statically analyzed, this looks good. If you have concerns with the update of the root password: Can you even log in as root if the password would have been updated/changed? We could still log the newly set password if you consider this a problem. Keep the security implications in mind, though.

christianlupus:
Ah, and we have to remember that this needs to be reversed once MySQL 9 is in use. But we will see that during tests.

scubamuc:
@Pilzinsel64 would this have any effect on "HPB client-push configuration" service configuration?

Pilzinsel64:
> would this have any effect on "HPB client-push configuration" service configuration?

No. It doesn't change the password, only how MySQL stores it, if i'm right.

pachulo (approved):

stondino00:
@pachulo this seems to work perfect.  Great work.

kyrofa:
> No. It doesn't change the password, only how MySQL stores it, if i'm right.

Correct.

pachulo:
So, time to merge! Thanks a lot for this @kyrofa !

Issue #2663, opened 2024-01-27
Should the snap offer the option to use APCu for local cache?


pachulo:
After reading the [Recommendations based on type of deployment](https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/caching_configuration.html#recommendations-based-on-type-of-deployment) on the [Memory caching page of the nextcloud manual](https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/caching_configuration.html), I was wondering:
* Should the snap come with the PHP APCu extension so that it can be used for `memcache.local` instead of Redis?

I haven't found any real benchmarks or anything, just that [it should be faster but consume more RAM](https://wpjohnny.com/redis-vs-memcache-vs-apcu-object-cache-comparisons/), but as it is also the official recommendation, I just wanted to understand better why we use Redis instead.

scubamuc:
@pachulo we have a large user base running "limited memory" devices (SBC's) so this should definitely be optional. 

> I haven't found any real benchmarks or anything, just that [it should be faster but consume more RAM](https://wpjohnny.com/redis-vs-memcache-vs-apcu-object-cache-comparisons/), but as it is also the official recommendation, I just wanted to understand better why we use Redis instead.

According to the recommendation, it could increase performance on "enhanced memory" machines. That would be desirable.

kyrofa:
Yeah, it makes sense that we'd be using redis given our start on the Nextcloud box. However, while I agree with @scubamuc that continuing to support such users is important, I just checked the metrics this morning: literally 95% of our users are amd64. If this would enhance performance, it might be worth investigating.

pachulo:
OK, I found more information:
* In [this nextcloud tunning guide from 2019](https://autoize.com/tuning-nextcloud-performance/) they talk about 2GB being the baseline for this:
  > If you are running on a small cloud instance (say, less than 2GB RAM), or a resource-constrained device such as Raspberry Pi, Redis can be used for both local caching and file locking.
* And they also reference [a (very old) benchmark from](https://github.com/owncloud-archive/pi-image/issues/24) @oparoz :
  > If you have the memory overhead to run both PHP APCu and Redis, you would enjoy better performance by [running APCu as the local cache, and Redis as the file locking cache](https://github.com/owncloud/documentation/pull/1992#issuecomment-203554416) based on a user benchmark on GitHub.


P.D.
[...]

PR #3449, opened 2026-04-19
php: add apcu extension

pachulo:
Let's see first if it's enough with this code to add the extension and manually configure the extension for local cache.

If that works, we can think about how to autoconfig this, based on available RAM.


pachulo:
I'm getting this error when building the snap:
```
Priming php 
+ snapcraftctl prime
[Errno 2] No such file or directory: '/root/stage/include/php/ext/apcu'
```

But the build of the extension itself seems to be fine:

```
Building php-ext-apcu 
+ /root/stage/bin/phpize
Configuring for:
PHP Api Version:         20230831
Zend Module Api No:      20230831
Zend Extension Api No:   420230831
+ ./configure
checking for grep that handles long lines and -e... /bin/grep
checking for egrep... /bin/grep -E

...

creating libtool
appending configuration tag "CXX" to libtool
configure: patching config.h.in
configure: creating ./config.status
config.status: creating config.h
+ sed -i.back 's|/parts/php/install/lib/php/extensions/|/parts/php-ext-apcu/install/lib/php/extensions/|g' Makefile
+ snapcraftctl build
<frozen importlib._bootstrap>:488: Warning: OpenSSL 3's legacy provider failed to load. Legacy algorithms will not be available. If you need those algorithms, check your OpenSSL configuration.
+ make -j4

...

(cd .libs && rm -f apcu.la && ln -s ../apcu.la apcu.la)
/bin/bash /root/parts/php-ext-apcu/build/libtool --tag=CC --mode=install cp ./apcu.la /root/parts/php-ext-apcu/build/modules
cp ./.libs/apcu.so /root/parts/php-ext-apcu/build/modules/apcu.so
cp ./.libs/apcu.lai /root/parts/php-ext-apcu/build/modules/apcu.la
PATH="$PATH:/sbin" ldconfig -n /root/parts/php-ext-apcu/build/modules
----------------------------------------------------------------------
Libraries have been installed in:
   /root/parts/php-ext-apcu/build/module

...

Build complete.
Don't forget to run 'make test'.

+ make -j4 install DESTDIR=/root/parts/php-ext-apcu/install
Installing shared extensions:     /root/parts/php-ext-apcu/install/lib/php/extensions/no-debug-non-zts-20230831/
Installing header files:          /root/parts/php/install/include/php/
```

Do you know what is wrong @christianlupus ?

christianlupus:
Just from your output, I directly see this line:

> ```
> Installing header files:          /root/parts/php/install/include/php/
> ```

I suspect/fear/assume there is more to it, to be honest. This would probably cause issues during runtime or compiling adjacent extensions if I understand correctly. Maybe one can fix this in one "simple" location, though. 

Just a quick analysis on the phone.

snap/snapcraft.yaml
```diff
@@ -304,6 +304,25 @@ parts:
...
+      # Prepare the php extension
+      ${SNAPCRAFT_STAGE}/bin/phpize
+
+      # Manually run the configure script
+      ./configure
+
+      # Patch the Makefile to install to the correct location in the snap
+      sed -i.back "s|/parts/php/install/lib/php/extensions/|/parts/php-ext-apcu/install/lib/php/extensions/|g" Makefile
+
+      snapcraftctl build
```
christianlupus, line 332: I read somewhere that it should be  `snapcraftctl default`, but this is without consulting the docs.

christianlupus:
One more point (and I did overlook this in redis case as well):

```
+ make -j4
/bin/bash /root/parts/php-ext-apcu/build/libtool --tag=CC --mode=compile cc -I. -I/root/parts/php-ext-apcu/build -I/root/parts/php-ext-apcu/build/include -I/root/parts/php-ext-apcu/build/main -I/root/parts/php-ext-apcu/build -I/root/parts/php/install/include/php -I/root/parts/php/install/include/php/main -I/root/parts/php/install/include/php/TSRM -I/root/parts/php/install/include/php/Zend -I/root/parts/php/install/include/php/ext -I/root/parts/php/install/include/php/ext/date/lib  -isystem/root/stage/include -DHAVE_CONFIG_H  -isystem/root/stage/include -D_GNU_SOURCE   -D_GNU_SOURCE -DZEND_ENABLE_STATIC_TSRMLS_CACHE=1 -Wall -Wextra -Wno-clobbered -Wno-unused-parameter -DZEND_COMPILE_DL_EXT=1 -c /root/parts/php-ext-apcu/build/apc.c -o apc.lo  -MMD -MF apc.dep -MT apc.lo
```

It tries to fetch headers from the PHP library's install folder. Instead, it should be the stage folder. Maybe that sed needs some more modifications 😉.

snap/snapcraft.yaml
pachulo: That [will be the case when we upgrade to core22](https://documentation.ubuntu.com/snapcraft/stable/how-to/change-bases/change-from-core20-to-core22/), (and also using `craftctl`  instead of `snapcraftctl`) but not just yet.

christianlupus:
Let's see if that will make it compile. Funny part is that on my dev machine, the build is not failing. So, I have to wait for the CI build.

pachulo:
Installed the build from this PR, but I don't see the PHP module in the serverinfo page:
image

And it's also not listed when trying this:
```
$ snap run --shell nextcloud.occ
$ ubuntu@nextcloud-test:/home/ubuntu$ /snap/nextcloud/current/bin/php --modules
[PHP Modules]
bcmath
bz2
Core
ctype
curl
date
dom
exif
fileinfo
filter
ftp
gd
gmp
hash
iconv
intl
json
ldap
libxml
mbstring
mysqlnd
openssl
pcntl
pcre
PDO
pdo_mysql
pdo_sqlite
Phar
posix
random
Reflection
session
SimpleXML
sodium
SPL
sqlite3
standard
sysvsem
tokenizer
xml
xmlreader
xmlwriter
zip
zlib

[Zend Modules]
```
But in this case, the redis one neither appears.

christianlupus:
I played around a bit and I think I found the issues.
When using `snap run --shell`, you do not use the correct `php.ini` by default. You can use
```
php -c $SNAP/config/php/php.ini -i
```
to see the current config. Optionally pipe through less fo better navigation.

Further, the PHP configuration was not loading the apcu extension. I, however, do not know if there are more settings needed in the PHP config to make use of the extension.

pachulo:
OK, it's working now!
After configuring it, the instance is working fine:
```
$ sudo nextcloud.occ -n config:system:set memcache.local --value="\OC\Memcache\APCu" --type=string
System config value memcache.local set to string \OC\Memcache\APCu
$ sudo nextcloud.occ config:list system | grep -b1 -a1  AP
462-        "memcache.locking": "\\OC\\Memcache\\Redis",
515:        "memcache.local": "\\OC\\Memcache\\APCu",
565-        "redis": {
```
And the php errors log is empty:
```
$ sudo cat /var/snap/nextcloud/53348/logs/php_errors.log
```

pachulo:
@nextcloud-snap/developers what do you think about creating an autoconfig for this, based on available RAM?

From what [I could find back in the day](https://github.com/nextcloud-snap/nextcloud-snap/issues/2663#issuecomment-1915646841), this should be on for local memcache on systems with >= 2GB.

The official documentation says that [APCu should be configured for the local memcache when on a "small/private home server"](https://docs.nextcloud.com/server/stable/admin_manual/configuration_server/caching_configuration.html#small-private-home-server) and that "[If you have enough memory, use APCu for Memory Caching and Redis for File Locking. If you are low on memory, use Redis for both.](https://docs.nextcloud.com/server/stable/admin_manual/configuration_server/caching_configuration.html#additional-notes-for-redis-vs-apcu-on-memory-caching)".
[Also](https://docs.nextcloud.com/server/stable/admin_manual/configuration_server/caching_configuration.html#additional-notes-for-redis-vs-apcu-on-memory-caching:~:text=Depending%20on%20your%20installation%20size%20and%20the%20number%20of%20users%20and%20interactions%20with%20the%20system%20you%20may%20want%20to%20adapt%20the%20apc.shm_size%20setting%20in%20your%20php.ini.%20The%20default%20value%20is%2032M%20which%20is%20usually%20too%20low%20for%20Nextcloud.%20A%20good%20starting%20point%20is%20128M.):
> Depending on your installation size and the number of users and interactions with the system you may want to adapt the apc.shm_size setting in your php.ini. The default value is 32M which is usually too low for Nextcloud. A good starting point is 128M. If you have a lot of users and/or a lot of apps installed you may want to increase this value further. Keep in mind that this memory needs to be available in your system’s memory and kept in mind when sizing the amount of workers on your server.

The other option would be to leave this as a manual knob, but I think that should be a last resort plan.
WDYT?

scubamuc:
@pachulo if we could auto-configure this it would be great... the manual knob is a last resort and we'll document any manual settings in the wiki making it available for folks needing it.

christianlupus:
Generally, I am with you. The snap addresses mostly private users and eventually small instances. No high availability, clustering, etc. Nevertheless, there is the point of very low-end devices (think raspberry pi) that rely on snap. So, I think a certain cutoff ram amount might be feasible. Adding a command or a config setting to enable (or disable) redis usage might be considerable for later fine-tuning.

scubamuc:
> Generally, I am with you. The snap addresses mostly private users and eventually small instances. No high availability, clustering, etc. Nevertheless, there is the point of very low-end devices (think raspberry pi) that rely on snap. So, I think a certain cutoff ram amount might be feasible. Adding a command or a config setting to enable (or disable) redis usage might be considerable for later fine-tuning.

as mentioned here https://github.com/nextcloud-snap/nextcloud-snap/issues/2663#issuecomment-1914868313 the metrics show 5% low powered SBC's (2024), important nonetheless! so if this is enabled by default for 95% and disabled when memory is below (say 2GiB) or even depending on arch (amd64 vs. arm64) may be feasible?

>  ...worth investigating as long as we can figure out a way for it to not hurt the other 5%.

snap/snapcraft.yaml
```diff
@@ -304,6 +304,33 @@ parts:
...
+      - php
+
+    override-build: |
+      # Prepare the php extension
+      ${SNAPCRAFT_STAGE}/bin/phpize
+
+      # Manually run the configure script
+      ./configure
+
+      # Patch the Makefile to install to the correct location in the snap
```
pachulo, line 321: @christianlupus should we do the same for the redis extension then?

pachulo:
OK, added some code to check the available RAM and, if it's equal or more than 2048 MB, set APCu as the local memcache. A simple test on a 32GB machine did work:
```
May 31 22:38:21 netxcloud-test nextcloud.apache[4955]: System config value memcache.locking set to string \OC\Memcache\Redis
May 31 22:38:22 netxcloud-test nextcloud.apache[5023]: Nextcloud or one of the apps require upgrade - only a limited number of commands are available
May 31 22:38:22 netxcloud-test nextcloud.apache[5023]: You may use your browser or the occ upgrade command to do the upgrade
May 31 22:38:22 netxcloud-test nextcloud.apache[5023]: System config value memcache.local set to string \OC\Memcache\APCu
```

I guess now I will need to try this on a container with limited RAM.

snap/snapcraft.yaml
christianlupus: We might consider it. Who knows what future redis extension updates will bring...

christianlupus:
Can you restrict the lxd containers on your machine? Otherwise, I could start up a limited VPS.

scubamuc:
you can change the memory and cpu allocations on LXC on the fly with:

https://documentation.ubuntu.com/lxd/latest/reference/instance_options/

`lxc config set <LXCNAME> limits.memory 2GB` 

for hard memory allocation use:

`lxc config set <LXCNAME> limits.memory.enforce hard`

check your config

`lxc config show LXCNAME`

```
image.version: "26.04"
  limits.memory: 2GB
  limits.memory.enforce: hard
  security.nesting: "true"
  user.comment: NC-SNAP Testing

```

scubamuc:
so I ran a quick test #3449 and can confirm the apcu warning is disappeared on system > 2GB, well done :heavy_check_mark: 

## Test fresh install > 2GB

grafik

grafik

scubamuc:
## Test fresh install < 1GB

* set LXC memory 1GB

```
lxc config set YCLOUD limits.memory 1GB
lxc config set YCLOUD limits.memory.enforce hard
```
* confirm memory

```
               total        used        free      shared  buff/cache   available
Mem:           953Mi       749Mi        57Mi       327Mi       473Mi       203Mi

```

* run install tests

grafik

grafik

grafik

looks like acpu is loaded still?

pachulo:
> looks like acpu is loaded still?

Yeah, that's expected, as right now we load the extension unconditionally, whether it's gonna be actually used for local memcache or not.
That means that 32MB (apc.shm_size default value) of RAM are gonna be allocated, even when not used by nextcloud. But then, if we unload it for systems with small RAM, a warning will appear on the admin portal... 🫠

scubamuc:
> > looks like acpu is loaded still?
> 
> Yeah, that's expected, as right now we load the module unconditionally, whether it's gonna be actually used for local memcache or not. That means that 32MB of RAM are gonna be allocated (apc.shm_size default value) even when not in use. But then, if we unload it for systems with small RAM, a warning will appear on the admin portal... 🫠

yes that makes perfect sense... :heavy_check_mark:

scubamuc (approved):

pachulo:
@kyrofa according to [the documentation](https://docs.nextcloud.com/server/stable/admin_manual/configuration_server/caching_configuration.html#apcu):

> Depending on your installation size and the number of users and interactions with the system you may want to adapt the apc.shm_size setting in your php.ini. The default value is 32M which is usually too low for Nextcloud. A good starting point is 128M. If you have a lot of users and/or a lot of apps installed you may want to increase this value further. Keep in mind that this memory needs to be available in your system’s memory and kept in mind when sizing the amount of workers on your server.

So, which options do you think we have to add a knob in the snap that allows users to modify the value of the `apc.shm_size` directive on the `php.ini` file?

Thanks!

pachulo:
OK, this is what I found out after looking for more information about the warning  "lack of apcu php module" that is new on nextcloud 33:
1. The module is required since upstream implemented https://github.com/nextcloud/server/pull/55728
2. They need it because "PHP-FPM and Apache module share memory and allow the use of apcu_inc to keep sequence ID."
3. Since this is now required by nextcloud core, I think that it makes sense to configure the memcache.local to use it for systems with >= 2GB of RAM.

The only problem I can think of is large instances where the default 32MB of the `apc.shm_size` won't be enough, and [nextcloud will show a warning on the admin page](https://docs.nextcloud.com/server/stable/admin_manual/configuration_server/caching_configuration.html#additional-notes-for-redis-vs-apcu-on-memory-caching:~:text=There%20is%20an%20admin%20check%20trying%20to%20detect%20too%20low%20memory%20sizing). In such case the snap doesn't offer a way to neither:
1.1 go back to Redis for `memcache.local`
1.2 configure `apc.shm_size` in `php.ini`

So, how you think we should proceed @nextcloud-snap/developers ?

scubamuc:
@pachulo 

> The only problem I can think of is large instances where the default 32MB of the apc.shm_size won't be enough, and [nextcloud will show a warning on the admin page](https://docs.nextcloud.com/server/stable/admin_manual/configuration_server/caching_configuration.html#additional-notes-for-redis-vs-apcu-on-memory-caching:~:text=There%20is%20an%20admin%20check%20trying%20to%20detect%20too%20low%20memory%20sizing)....

* can you be clear, how big is "large" = how many users?
* so this is a one-way ticket?

pachulo:
> can you be clear, how big is "large" = how many users?

I have no idea; I'm just parroting [the documentation](https://docs.nextcloud.com/server/stable/admin_manual/configuration_server/caching_configuration.html#apcu):
> A good starting point is 128M. If you have a lot of users and/or a lot of apps installed you may want to increase this value further.


> so this is a one-way ticket?

@scubamuc I don't get what you mean here, sorry.

Anyway, I think that we should:
1. Strip this PR, so that it only adds the apcu extension.
2. Open another issue about `memcache.local`, and continue the discussion there about whether to autoconfigure APCu and how we can grow a knob to allow for `apc.shm_size` configuration.

This plan should allow us to release 33.0.5, as the warning will be gone.

scubamuc:
> This plan should allow us to release 33.0.5, as the warning will be gone

:heavy_check_mark: :tada:

christianlupus:
I think APCu is suggested for local memcache in any case (in the [docs](https://docs.nextcloud.com/server/stable/admin_manual/configuration_server/caching_configuration.html#recommendations-based-on-type-of-deployment)), I think this should be done in any case.
Adding more configs should be managable anyways. Just open an issue to keep it in mind.

scubamuc (approved):

Pilzinsel64 (approved):
👍

PR #3452, opened 2026-04-19
snap: disable armhf builds

pachulo:


kyrofa (approved):
This is too bad, but we don't really have much of a choice I suppose.

Pilzinsel64 (approved):

Issue #1902, opened 2021-10-13
Drop certbot in favor of something with less dependencies


kyrofa:
I recommend [acme.sh](https://github.com/acmesh-official/acme.sh) or something similar. Requirements:

- Existing installations shouldn't even notice the switch (will require configuration [migrations](https://github.com/nextcloud-snap/nextcloud-snap/tree/master/src/migrations))
  - Existing certs should continue renewing seamlessly (see the [`renew-certs`](https://github.com/nextcloud-snap/nextcloud-snap/blob/master/src/https/bin/renew-certs) executable)
- Existing CLI commands and configuration parameters should not change API
  - I believe this is only the [`enable-https`](https://github.com/nextcloud-snap/nextcloud-snap/blob/master/src/https/bin/enable-https) executable

r4co0n:
Nice, there is a migrations construct available, this should make things easy. I am currently having my eyes on [mod_md](https://github.com/icing/mod_md/), which might integrate certificate generation into normal webserver operations. acme.sh is another option I'm evaluating.

kyrofa:
That's funny, mod_md caught my eye as well. My own concern was that it was apache specific, which would get really annoying if we ever wanted to move to, say, nginx. No strong feelings though, explore.

r4co0n:
When thinking about how to migrate our users to another mechanism for providing letsencrypt or potentially any certificates from an ACME-capable authority - Scott Helme has some nice blog posts about those, [here](https://scotthelme.co.uk/heres-another-free-ca-as-an-alternative-to-lets-encrypt/) is the latest, including links to previous discussions of other providers - the hurdle of identifying the currently selected encryption method for the snap came up in #1924, quoting myself:
> The biggest concern with migration is keeping everything working for all the folks that currently use certbot, and for those that have provided their own certificates. As far as I could discern, we never log what people last chose when setting up encryption, or am I mistaken? It's prudent to know if the current certificate is self-signed (generated by us), provided by letsencrypt or provided by the admin.

Is there a sane way to discern what was select as certificate source when our user last called `nextcloud.enable-https`?

Keep in mind that users could have been doing the switcheroo and used every possible configuration in the past, with certificates that have not yet expired. And there might well be users providing LetsEncrypt certificates using the custom mechanism, as they fetch them another way.

kyrofa:
> It's prudent to know if the current certificate is self-signed (generated by us), provided by letsencrypt or provided by the admin.
[...]

PR #3456, opened 2026-04-20
Migrate legacy webroot plugin

### LINKED ISSUE #3419: Drop nextcloud plugin for certbot
We already discussed this, but I found no matching issue, thus I create this one to keep track of the progress:

We have the ancient (and outdated) certbot nextcloud (custom) plugin in some old instances in place. We should prior to #1902 try to reduce the number of variants out in the wild. By migrating the nextcloud plugin to the current webroot plugin, we can rule one system out over the time.

christianlupus:
This is a fixer script to help get rid of the manually managed Nextcloud plugin for Certbot. 

This one needs thorough testing on an instance migrated from the ancient times to check the migration script actually works as expected.

This is a prestep for #1902 
Prepares #3419

---

- [x] Use current webroot plugin to obtain a valid certificate. The fixer should not try to create new certs.
- [x] Use of current webroot cert from staging should not trigger a new certificate.
- [x] Usage of self-signed certs should not trigger a new certificate
- [x] Usage of custom certs should not trigger a new certificate
- [x] Usage of a legacy staging certificate should migrate to a webroot staging certificate (using 1 domain)
- [x] Usage of a legacy staging certificate should migrate to a webroot staging certificate (using multiple domains)
- [x] Usage of a legacy production certificate should migrate to a webroot production certificate

### DIFF
--- a/src/nextcloud/fixes/existing-install/maintenance/04_migrate-webroot-plugin.sh
+++ b/src/nextcloud/fixes/existing-install/maintenance/04_migrate-webroot-plugin.sh
@@ -1,0 +1,142 @@
+#!/bin/sh
+# 
+# The custom nextcloud plugin for certbot was replaced by the official 
+# webroot plugin in Version 15.0.8snap2. This script should migrate
+# existing installations to use the webroot plugin.
+
+set -e
+
+# shellcheck source=src/https/utilities/https-utilities
+. "$SNAP/utilities/https-utilities"
+
+printf "Checking if the certbot plugin needs to be migrated... "
+
+if ! certificates_are_active; then
+    echo "no (no HTTPS certificates are active)"
+    exit
+fi
+
+if self_signed_certificates_are_active || custom_certificates_are_active; then
+    echo "Not using certbot certificates, so no migration is needed."
+    exit
+fi
+
+certdir="$(get_most_recent_certificate_directory)"
+RENEWAL_CONFIG_FILE="$SNAP_CURRENT/certs/certbot/config/renewal/$certdir.conf"
+
+if [ ! -f "$RENEWAL_CONFIG_FILE" ]; then
+    echo "no (renewal config file not found)"
+    exit 1
+fi
+
+if grep -q "authenticator = nextcloud:webroot" "$RENEWAL_CONFIG_FILE"; then
+    echo "yes"
+else
+    echo "no"
+    exit 0
+fi
+
+# Extract the domains from the certificate
+AWK_CMD="/^[[:space:]]*X509v3 Subject Alternative Name:/ {
+    capture=1
+    next
+}
+
+/^[[:space:]]*X509v3/ {
+    if (capture) exit
+}
+
+capture {
+    print
+}"
+
+domains=$(openssl x509 -noout -text -in "$CERTBOT_LIVE_DIRECTORY/$certdir/cert.pem" | mawk "$AWK_CMD" | sed -e 's/^[[:space:]]*//g' | tr -d "," | sed -e 's/DNS://g')
+echo "Found domains to be migrated: $domains"
+
+# Collect all parameters to be used with certbot.
+extra_params=""
+
+# Check the current ACME server
+acme_server=$(grep -E "^server\s*=\s*" "$RENEWAL_CONFIG_FILE" | sed -e 's/^[[:space:]]*server\s*=\s*//g')
+if [ "$acme_server" = "https://acme-v02.api.letsencrypt.org/directory" ]; then
+    # Default (productive) ACME server, so nothing is needed
+    :
+elif [ "$acme_server" = "https://acme-staging-v02.api.letsencrypt.org/directory" ]; then
+    # Staging ACME server, so we need to add the --staging flag to the certbot command
+    extra_params="--staging"
+else
+    echo "error: unrecognized ACME server: $acme_server" >&2
+    exit 1
+fi
+
+for domain in $domains; do
+    extra_params="$extra_params -d $domain"
+done
+
+printf "Testing if certificates can be obtained with the webroot plugin... "
+
+# Building CLI commands, so we don't WANT to quote some of these (they need
+# to be separated by whitespace): disable the check
+# shellcheck disable=SC2086
+if output="$(run_certbot_certonly $extra_params --staging --dry-run 2>&1)"; then
+    echo "success"
+else
+    echo "failed!"
+    echo "error running certbot:" >&2
+    echo "" >&2
+    echo "$output" >&2
+    exit 1
+fi
+
+# Copy the legacy plugin to a temporary location.
+cp -R --preserve=timestamps,mode "$SNAP_CURRENT/certs/certbot" "$SNAP_CURRENT/certs/certbot.legacy"
+
+printf "Migrating certbot configuration to use the webroot plugin... "
+
+trap_for_error() {
+    retVal=$?
+
+    # No issue found. We are safe. Exit with success.
+    if [ "$retVal" -eq 0 ]; then
+        exit 0
+    fi
+
+    echo ""
+    echo "An error occurred during the migration process. Restoring legacy certbot configuration."
+
+    rm -rf "$SNAP_CURRENT/certs/certbot"
+    mv "$SNAP_CURRENT/certs/certbot.legacy" "$SNAP_CURRENT/certs/certbot"
+
+    exit "$retVal"
+}
+# Install a trap in case something goes wrong during the migration process, so we can restore the
+# legacy configuration and avoid leaving the user with a broken certbot configuration.
+trap trap_for_error EXIT INT TERM HUP
+
+# Drop old configuration files of the active certificate
+rm -rf \
+    "$SNAP_CURRENT/certs/certbot/config/archive/$certdir" \
+    "$SNAP_CURRENT/certs/certbot/config/live/$certdir" \
+    "$SNAP_CURRENT/certs/certbot/config/renewal/$certdir.conf"
+
+# Building CLI commands, so we don't WANT to quote some of these (they need
+# to be separated by whitespace): disable the check
+# shellcheck disable=SC2086
+if output="$(run_certbot_certonly $extra_params 2>&1)"; then
+    echo "success"
+
+    # Disable the trap to avoid restoring the legacy configuration after a successful migration
+    trap - EXIT INT TERM HUP
+else
+    echo "failed!"
+    echo "error running certbot:" >&2
+    echo "" >&2
+    echo "$output" >&2
+
+    exit 1
+fi
+
+echo "Removing legacy certbot configuration."
+rm -rf "$SNAP_CURRENT/certs/certbot.legacy"
+
+activate_certbot_certificate
--- a/src/nextcloud/fixes/existing-install/maintenance/4_migrate-webroot-plugin.sh
+++ b/src/nextcloud/fixes/existing-install/maintenance/4_migrate-webroot-plugin.sh
@@ -1,0 +1,104 @@
+#!/bin/sh
+# 
+# The custom nextcloud plugin for certbot was replaced by the official 
+# webroot plugin in Version 15.0.8snap2. This script should migrate
+# existing installations to use the webroot plugin.
+
+set -e
+
+# shellcheck source=src/https/utilities/https-utilities
+. "$SNAP/utilities/https-utilities"
+
+echo -n "Checking if the certbot plugin needs to be migrated... "
+
+if ! certificates_are_active; then
+    echo "no (no HTTPS certificates are active)"
+    exit
+fi
+
+certdir="$(get_most_recent_certificate_directory)"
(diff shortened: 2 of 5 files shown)


scubamuc:
> This one needs thorough testing on an instance migrated from the ancient times to check the migration script actually works as expected.

thorough testing impossible with terminating rproxy in place. @nextcloud-snap/developers any suggestions? 

call for testing?

kyrofa:
You can still install ancient snaps by using channels, of course (e.g. `snap install nextcloud --channel=14`), but whether or not that's helpful will depend on that old certbot's ability to actually obtain certs. I'm not entirely sure it'll work anymore. If it does, though, that's the ideal way to test this.

src/nextcloud/fixes/existing-install/maintenance/4_migrate-webroot-plugin.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/sh
```
kyrofa, line 1: I'm not sure the fixer is the right place for this script, especially not in the maintenance mode section of it (ideally that is as short as possible). Does this need services running? If not, a migration might be better suited (which is run via the post-install hook).

scubamuc:
wow... don't you love this:

nextcloud (14/stable) 14.0.10snap1 from Nextcloud✓ installed

grafik

best since sliced toast :open_mouth: the auto-installer works! :+1: 

grafik

src/nextcloud/fixes/existing-install/maintenance/4_migrate-webroot-plugin.sh
christianlupus: I was thinking about 3 possible locations: the fixer, a migration, and the renew script. Each has their own benefits and drawbacks. 

What I need at minimum is a running apache server if if do not misunderstand this setup completely. Nextcloud, database, and other stuff can still be offline or in whatever state. I decided for maintenance as it might interrupt network connection for the restart of the apache server. 

I could migrate this script from the fixer to the maintenance section. I have to check if the prior running services will be up/starting while the Post-Upgrade aka migrations hook is executed. I feared that apache would not be there on a guaranteed level. For the fixer it is. 

Using the renew script is the guaranteed version that in 90 days there are no legit certificates with the old plugin or there. However the renew is not the right place to do migrations in my understanding. Also this will cause trouble in the future if something went wrong. So I voted against this option (mentally).

kyrofa: Given that we're running certbot standalone (not using any sort of Apache integration), why does Apache need to be running here?

christianlupus:
In #3419 I did exactly this: install NC 14 (which used the old plugin by default) and extracted the data for later inspection. It is linked there just for reference. 

Currently, there is I think one or two bugs in the implementation. So, please let me fix them prior to putting too much time into this.

src/nextcloud/fixes/existing-install/maintenance/4_migrate-webroot-plugin.sh
```diff
@@ -0,0 +1,104 @@
...
+    echo "error running certbot:\n\n" >&2
+    echo "$output" >&2
+
+    echo "Restoring legacy certbot configuration."
+    rm -rf "$SNAP_CURRENT/certs/certbot"
+    mv "$SNAP_CURRENT/certs/certbot.legacy" "$SNAP_CURRENT/certs/certbot"
+
+    exit 1
+fi
+
```
christianlupus, line 119: Apache is not restarted in case of successful obtaining a new certificate.

src/nextcloud/fixes/existing-install/maintenance/4_migrate-webroot-plugin.sh
```diff
@@ -0,0 +1,104 @@
...
+for domain in $domains; do
+    extra_params="$extra_params -d $domain"
+done
+
+echo -n "Testing if certificates can be obtained with the webroot plugin... "
+
+if output="$(run_certbot_certonly $extra_params --staging --dry-run" 2>&1)"; then
```
christianlupus:
```suggestion
if output="$(run_certbot_certonly $extra_params --staging --dry-run 2>&1)"; then
```

christianlupus (commented):
Bugs to be fixed

src/nextcloud/fixes/existing-install/maintenance/4_migrate-webroot-plugin.sh
christianlupus: The certbot just handles the acme part. The server will then pose a challenge to be solved. To do so, certbot must make a random file accessible at a dedicated location. Technically, it is just placed into a folder (thi6s the name webroot of that plugin) where any http server will pick the file as a static one up and serve it. This apache will be needed to fulfill the challenge (at least the http-01). See the [config](https://github.com/nextcloud-snap/nextcloud-snap/blob/d35bf9e11a5739fe6f5692c3c46e87fe530de7e8/src/apache/conf/httpd.conf#L142).

kyrofa: Ah yes, thank you, that was the piece I was missing. You're right, this is really the only place then. I so wish snapd would give us better hooks.

src/nextcloud/fixes/existing-install/maintenance/4_migrate-webroot-plugin.sh
```diff
@@ -0,0 +1,123 @@
...
+    rm -rf "$SNAP_CURRENT/certs/certbot"
+    mv "$SNAP_CURRENT/certs/certbot.legacy" "$SNAP_CURRENT/certs/certbot"
+
+    exit 1
+fi
+
+echo "Removing legacy certbot configuration."
+rm -rf "$SNAP_CURRENT/certs/certbot.legacy"
+
+activate_certbot_certificate
```
christianlupus, line 123: Currently, the restart (and replacement using `mv` & `rm`) is done unconditinally. If there was a staging cert in place, I assume, this is fine?

christianlupus:
Just thinking of some test to carry out.
I put them into the PR description.

Further questions to clarify:

1. Do/did we use at any point in time non-HTTP-01 challenges? That is DNS-01 or TLS-ALPN-01?
2. Are there more edge cases to be considered?

We can adjust this list as as do the tests or find further issues/concerns.

src/nextcloud/fixes/existing-install/maintenance/4_migrate-webroot-plugin.sh
```diff
@@ -0,0 +1,123 @@
...
+
+# Check the current ACME server
+acme_server=$(grep -E "^server\s*=\s*" "$RENEWAL_CONFIG_FILE" | sed -e 's/^[[:space:]]*server\s*=\s*//g')
+if [ "$acme_server" = "https://acme-v02.api.letsencrypt.org/directory" ]; then
+    # Default (productive) ACME server, so nothing is needed
+    :
+elif [ "$acme_server" = "https://acme-staging-v02.api.letsencrypt.org/directory" ]; then
+    # Staging ACME server, so we need to add the --staging flag to the certbot command
+    extra_params="--staging"
+    dry_run=y
```
kyrofa, line 68: What is `dry_run` for? Looks like it just results in `--staging` being added, which isn't really a dry run, and also seems redundant given what we're doing here with `extra_params`, no? 🤔

src/nextcloud/fixes/existing-install/maintenance/4_migrate-webroot-plugin.sh
```diff
@@ -0,0 +1,123 @@
...
+    echo "success"
+else
+    echo "failed!"
+    echo "error running certbot:" >&2
+    echo "" >&2
+    echo "$output" >&2
+
+    echo "Restoring legacy certbot configuration."
+    rm -rf "$SNAP_CURRENT/certs/certbot"
+    mv "$SNAP_CURRENT/certs/certbot.legacy" "$SNAP_CURRENT/certs/certbot"
```
kyrofa, line 115: Given the use of `set -e`, would it be better to have this in a `trap`?

src/nextcloud/fixes/existing-install/maintenance/4_migrate-webroot-plugin.sh
kyrofa: I'm not sure I'm following you here. If we got to this point, we know we fetched new certs, right? So a restart is always going to be required at this point, is it not?

src/nextcloud/fixes/existing-install/maintenance/04_migrate-webroot-plugin.sh
```diff
@@ -0,0 +1,123 @@
...
+    rm -rf "$SNAP_CURRENT/certs/certbot"
+    mv "$SNAP_CURRENT/certs/certbot.legacy" "$SNAP_CURRENT/certs/certbot"
+
+    exit 1
+fi
+
+echo "Removing legacy certbot configuration."
+rm -rf "$SNAP_CURRENT/certs/certbot.legacy"
+
+activate_certbot_certificate
```
christianlupus: I think we can ignore my comment. I was thinking this way:

I check the URL of the previously existing certificate. This one might theoretically be a staging cert. I was not sure if we would have to handle this separately. But thinking about it once more, I doubt it. If the user/admin wants to have staging certs, well, that's his decision to make. So, I would resolve this discussion.

src/nextcloud/fixes/existing-install/maintenance/4_migrate-webroot-plugin.sh
christianlupus: I dropped it. It was just crap. I thought I might need to handle staging certs differently.

src/nextcloud/fixes/existing-install/maintenance/4_migrate-webroot-plugin.sh
christianlupus: I thought about a trap as well to be honest.

I decided against it as only one single function call (`run_certbot_certonly`) needs protection. Further, it is not directly possible to collect the command output of the certbot command for `echo`ing it as part of the debug output. I instead went for a _mixed_ variant. I hope this is what you had in mind.

christianlupus:
I tried the following tests on a VPS and found one bug. I fixed it on this PR and trested a live patched version (using `snap try`). I would wait for the actual build before doing the update chain (14 up to 32).

- [x] (live patched) Use of current webroot cert from staging should not trigger a new certificate.
- [x] (live patched) Usage of self-signed certs should not trigger a new certificate
- [x] (live patched) Usage of custom certs should not trigger a new certificate

I want to redo these tests once the patch is completely incorporated.

christianlupus:
Yesterday, I started testing using this [script](https://github.com/christianlupus-nextcloud/nextcloud-snap/blob/webroot-plugin-data/tmp/3456-webroot/bashscripts/migrate-webroot.sh). So far, I ran a few tests successfully.

- [x] Use current webroot plugin to obtain a valid certificate. The fixer should not try to create new certs.
- [x] Use of current webroot cert from staging should not trigger a new certificate.
- [x] Usage of self-signed certs should not trigger a new certificate
- [x] Usage of custom certs should not trigger a new certificate
- [x] Usage of a legacy staging certificate should migrate to a webroot staging certificate (using 1 domain)
- [x] Usage of a legacy staging certificate should migrate to a webroot staging certificate (using multiple domains)
- [x] Usage of a legacy production certificate should migrate to a webroot production certificate

scubamuc (approved):

christianlupus:
Finally, I managed to check it. I was able to use this PR's snap'ed build to successfully migrate the ancient (pre-NC-15 installations) nextcloud:webroot plugin to the current webroot plugin. Also, current installations are not affected by this. There might be some corner cases with ancient installations but these will be very rare and in the worst case, a new certificate needs to be obtained.

What are the next steps? Do you want to check this in some way?
I would then rebase and squash this once we are okay with the content of this PR. Any further thoughts on this, @nextcloud-snap/developers?

pachulo:
Wow! First of all: thanks a lot for this @christianlupus ! 👏 

> What are the next steps? Do you want to check this in some way?

IMHO, your checks and tests are more than enough.

Also, I totally agree with what you said here:

> in the worst case, a new certificate needs to be obtained.

So, again IMHO, we are good to go with this one.

pachulo (approved):

christianlupus:
@kyrofa given the fact you had an unresolved comment on this, I would wait a bit for actually merging this PR. I would wait a bit, so if you want to have a closer look, please give me a quick ping. I will then delay this a bit.

src/nextcloud/fixes/existing-install/maintenance/4_migrate-webroot-plugin.sh
kyrofa: You've considered this, that's all I wanted, don't hold up on me!

kyrofa:
Nah, don't hold it up on me, you've covered your bases here.

Issue #3461, opened 2026-04-24
Certbot is broken in Nextcloud snap 32.0.8snap3 on ARM64.


rugeps:
#### Describe the bug
Certbot is broken in Nextcloud snap 32.0.8snap3 on ARM64.

Running Let's Encrypt certificate generation or renewal fails due to missing Python dependencies (zope.interface), and certbot itself cannot be executed.

#### To Reproduce

1. Run:

`sudo nextcloud.enable-https lets-encrypt`

2. Or run certbot directly:

`sudo /snap/nextcloud/current/bin/certbot --version`

3. Observe failure

Expected behavior:
Certbot should be correctly bundled and allow certificate generation and renewal without errors.

#### Screenshots

Not applicable (CLI error)

#### OS/snapd/snap version
```
$ snap list nextcloud
Name       Version      Rev    Tracking       Publisher   Notes
nextcloud  32.0.8snap3  53087  latest/stable  nextcloud✓  -

$ snap version
snap          2.75.2
snapd         2.75.2
series        16
ubuntu        22.04
kernel        5.10.110-rockchip-rk3588
architecture  arm64
```

#### Logs
```
#### Certbot failure via Nextcloud command
pkg_resources.DistributionNotFound: The 'zope.interface' distribution was not found

#### Direct certbot execution
ModuleNotFoundError: No module named 'certbot'

#### Full output

orangepi@orangepi5plus:~$ sudo nextcloud.enable-https lets-encrypt
Attempting to obtain certificates... error running certbot:

An unexpected error occurred:
pkg_resources.DistributionNotFound: The 'zope.interface' distribution was not found and is required by the application

orangepi@orangepi5plus:~$ sudo /snap/nextcloud/current/bin/certbot --version
Traceback (most recent call last):
  File "/snap/nextcloud/current/bin/certbot", line 5, in <module>
    from certbot.main import main
ModuleNotFoundError: No module named 'certbot'
```

Additional context
- Existing Let's Encrypt certificate is still valid, but renewal cannot be triggered manually
- Ports 80/443 are correctly forwarded and were working previously
- Issue appeared after a recent snap refresh (April 2026)
- Likely a packaging/runtime issue affecting certbot within the snap
- 

[output from the script.txt](https://github.com/user-attachments/files/27068907/output.from.the.script.txt)

Device: Orange Pi 5 Plus (ARM64)

scubamuc:
Hey @rugeps welcome and thanks for your request 👋 

[...]

PR #3464, opened 2026-04-28
certbot: add workaround for zope* dependencies on arm64

pachulo:
It looks like we are currently in a "messy middle" phase regarding Python metadata.

According to Gemini, on amd64 the building is likely pulling pre-compiled Wheels from PyPI. Those Wheels were built with an older naming convention, so they keep the dot: zope.*
On arm64, there likely isn't a pre-compiled Wheel for zope.* dependencies, it needs to be built from the Source Distribution (sdist). Because the build is happening on a newer core20 system, the modern build tools normalize the names to zope_*

Instead of fighting the naming convention, we simply "alias" the directories so that pkg_resources finds them on arm64.

I've built this on an arm64 VM and I've done this after installing it:
```
$ snap run --shell nextcloud.occ
To run a command as administrator (user "root"), use "sudo <command>".
See "man sudo_root" for details.

ubuntu@gleaming-cuckoo:/home/ubuntu$ certbot --version
certbot 0.39.0
```

So it seems to do the trick.

This should fix https://github.com/nextcloud-snap/nextcloud-snap/issues/3461

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -432,6 +432,22 @@ parts:
       snapcraftctl build
       patch -p1 -d $SNAPCRAFT_PART_INSTALL/lib/python3*/site-packages/certbot < $SNAPCRAFT_STAGE/certbot-remove-default-config-files.patch
       patch -p1 -d $SNAPCRAFT_PART_INSTALL/lib/python3*/site-packages/certbot < $SNAPCRAFT_STAGE/certbot-remove-storage-chown.patch
+    override-prime: |
+      snapcraftctl prime
+      cd "$SNAPCRAFT_PRIME/lib/python3.8/site-packages/"
+      
+      # If we have the dir with the underscore, make a link with the dot.
+      # This is needed to solve a issue affecting arm64. See #3461
+      for pkg in zope_interface zope_hookable zope_proxy
+      do
+        for filename in ./$pkg-*.dist-info
+        do
+          if [ -d "$filename" ]; then
+            # Replace underscores in pkg name, then strip the ./pkg prefix from filename
+            ln -sf "$filename" "${pkg//_/.}${filename#./$pkg}"
+          fi
+        done
+      done
 
   setup-https:
     plugin: dump


snap/snapcraft.yaml
```diff
@@ -432,6 +432,16 @@ parts:
...
+    override-prime: |
+      snapcraftctl prime
+      cd "$SNAPCRAFT_PRIME/lib/python3.8/site-packages/"
+      
+      # If we have the dir with the underscore, make a link with the dot.
+      # This is needed to solve a issue affecting arm64. See #3461
+      [ -d zope_interface-4.7.2.dist-info ] && ln -sf zope_interface-4.7.2.dist-info zope.interface-4.7.2.dist-info
+      [ -d zope_hookable-5.0.1.dist-info ] && ln -sf zope_hookable-5.0.1.dist-info zope.hookable-5.0.1.dist-info
+      [ -d zope_proxy-4.3.5.dist-info ] && ln -sf zope_proxy-4.3.5.dist-info zope.proxy-4.3.5.dist-info
+      true # Ensure the script returns 0 even if directories aren't found
```
christianlupus:
```suggestion
      for pkg in zope_interface zope_hookable zope_proxy
      do
        # Just unglob the filename
        for filename in ./$pkg-*.dist-info
        do
          if [ -d "$filename" ]; then
            ln -sf "$filename" "$(echo $pkg | sed '/_/./g')${filename#$pkg}"
          fi
        done
      done
```

Something in this way would make it independent of the exact version installed and also should avoid the explicit `true` as a fallback (hopefully abort in case of problems).

christianlupus (approved):
I cannot test here locally for the created artifact as I only have a very low-power orange pi with 1GB of RAM here. Building might take days.

Static analysis seems fine as discussed in the corresponding bug report.

I have just one remark: Maybe the actual linking might be made more robust to avoid issues not bubbling up.

snap/snapcraft.yaml
pachulo: I considered trying to find a more robust solution, but I ultimately feel that a "quick and dirty" hardcoded version might be better for now; it serves as a reminder that we need to replace it as soon as possible.
WDYT, @kyrofa? (Especially considering that @christianlupus’s solution is much more technically sound)

kyrofa: > it serves as a reminder that we need to replace it as soon as possible

Do we actually have a plan for doing so? This is a hack either way, being able to update the requirements.txt without also needing to remember to update the hard-coded versions here might be nice unless we literally have a plan for ripping this out before such a thing might happen.

pachulo: > Do we actually have a plan for doing so?

From what I understood, though I could very well be wrong, updating certbot should help us here.
For example, [this change from version 2.0.0](https://github.com/certbot/certbot/blob/main/certbot/CHANGELOG.md#200---2022-11-21) looks promising:
> Certbot no longer depends on zope.

And there is also [this change in version 2.8.0](https://github.com/certbot/certbot/blob/main/certbot/CHANGELOG.md#280---2023-12-05):
> Stop using the deprecated pkg_resources API included in setuptools.

kyrofa: Okay, what is holding us back from doing that at this point?

pachulo: That we first need to unblock the issue for arm64 users.
After that we can try the upgrade to the latest certbot that still supports python 3.8 ([3.0.1](https://github.com/certbot/certbot/blob/main/certbot/CHANGELOG.md#310---2025-01-07)), and properly test if everything is fine.

christianlupus: Looking at the packed certbot versions in Ubuntu 22.04 and 24.04, they are behind certbot v3. So, although it might be doable, we would have to be super careful here. I am unsure how far we can push certbot over the default versions (as present in the corresponding Ubuntu versions).

scubamuc:
nextcloud (beta/pr-3464) 0+git.0bb3c68 from Nextcloud✓ installed

```
Name       Version        Rev    Tracking       Publisher   Notes
core20     20260211       2769   latest/stable  canonical✓  base
nextcloud  0+git.0bb3c68  53180  latest/beta/…  nextcloud✓  -
snapd      2.75.2         26865  latest/stable  canonical✓  snapd
```

grafik

grafik

grafik

grafik

LGTM, WDYT?

scubamuc (approved):

christianlupus:
I would spin up a VPS tomorrow to test let's encrypt on amd at least.

snap/snapcraft.yaml
pachulo: @christianlupus & @kyrofa you've convinced me. Let's open an issue so we don't forget to remove this as soon as we're able.

pachulo:
OK, I've compiled this for arm64, with a modification of the changes that @christianlupus suggested, and everything looks fine:
```
$ snap run --shell nextcloud.occ
To run a command as administrator (user "root"), use "sudo <command>".
See "man sudo_root" for details.

ubuntu@gleaming-cuckoo:/home/ubuntu/nextcloud-snap$ certbot --version
certbot 0.39.0
```

I will be merging this, so that @rugeps can test it, once a build is uploaded to the `latest/beta` channel.

PR #3497, opened 2026-05-31
Release 32.0.10snap1

pachulo:


christianlupus (approved):

Pilzinsel64:
Problem: Push notifications are completely broken. If you have Notifications as only 2FA provider, you will not be able to login again until the next update.
https://github.com/nextcloud/notifications/issues/3035

Again the question: Should be ship anyway or ignore this fact? As this also affects login and "Notification" is a commonly used 2FA, I suggest to NOT ship this (again) and wait for 32.0.11.

pachulo:
> If you have Notifications as only 2FA provider, you will not be able to login again until the next update.

I'm not using this feature myself @Pilzinsel64 : do you have any idea of how "default" is to have the notifications as the only 2FA provider?

> Again the question: Should be ship anyway or ignore this fact? As this also affects login and "Notification" is a commonly used 2FA, I suggest to NOT ship this (again) and wait for 32.0.11.

I'm starting to grow tired of this QA-kind-of issues, but I have to agree with you: leaving, even some of our users, without the possibility to login, is a no-go.
I just hope upstream thinks the same and they make an out-of-band release 🙏

Pilzinsel64:
@pachulo Well, notifications is a default app which is shipped with Nextcloud. Also 2FA is recommended by Nextcloud for all users and should be used by default. So, using notifications as 2FA provider can be expected as common.

Also, notifications means also Talk. If there are instances that relay on this (small business instances or several home users that use talk as primary messenger) they will have issues. Talk notifications and also Files notifications are affected by this. Basically all Android apps. Desktop apps still work probably.

Pilzinsel64:
Closing in favor of #3504

PR #3523, opened 2026-06-14
Configure proxy buffering for PHP-FPM backend

### LINKED ISSUE #3522: Cannot disable PHP-FPM proxy buffering
In the Apache webserver configuration, the reverse proxy to the PHP-FPM backend is hardcoded such that proxy buffering is enabled ([source](https://github.com/nextcloud-snap/nextcloud-snap/blob/cfe6ec2a6189dd8e1729febc8e89e6e9bc005298/src/apache/conf/httpd.conf#L163)).  This breaks some services like [NextPush](https://codeberg.org/NextPush/uppush) that require proxy buffering to be disabled.

It would thus be helpful to allow disabling proxy buffering.

As a workaround, it is possible to use a bind mount to replace the snap's `http.conf` with a custom file that applies the required attributes on the reverse proxy config:

- Before: `<Proxy "fcgi://localhost/">`
- After: `<Proxy "fcgi://localhost/" disablereuse=on flushpackets=on max=10>`

AndrewFerr:
Fixes #3522

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,18 @@ To disable it, run:
     $ sudo snap set nextcloud http.compression=false
 
 
+#### PHP-FPM proxy buffering configuration
+
+By default, Apache's reverse proxy to PHP-FPM will not auto-flush output after each chunk of data.
+If you need the proxy to be configured without buffering, run:
+
+    $ sudo snap set nextcloud http.fcgi-flushpackets=true
+
+To disable it, run:
+
+    $ sudo snap set nextcloud http.fcgi-flushpackets=false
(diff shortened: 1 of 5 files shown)


AndrewFerr:
This applies the config change recommended by [the NextPush server](https://codeberg.org/NextPush/uppush#installation) to disable proxy buffering to the PHP-FPM backend.

Ideally this change would be configurable instead of hardcoded as it is now, but I'm not yet aware of how to do that, or what would be the correct component to control this config, so I am open to suggestions.

PR #3526, opened 2026-06-15
README: nextcloud snap is community driven project

pachulo:


christianlupus (approved):

README.md
Pilzinsel64, line 3:
```suggestion
Community-supported Nextcloud snap package, which consists of:
```

Or, alternatively:
```suggestion
Community-driven Nextcloud snap package, which consists of:
```

What do you think?

scubamuc: I don't understand why we don't simply use the same text that can be found in all the docs that are out there... the idea is to recognise and immediately understand that the text has one origin.

Pilzinsel64: It was just an idea, personal preference, think it sounds/reads better. But didn't think about existing docs, might be to late, sorry.

Pilzinsel64 (approved):

README.md
pachulo: > I don't understand why we don't simply use the same text that can be found in all the docs that are out there...

I've modified it to make it more similar, but I swapped `snap installation` with `project`, as I think it better fits the README context. WDYT @scubamuc ?

PR #3544, opened 2026-07-03
Update PHP to 8.4.23

### LINKED ISSUE #3487: php: bump to 8.4
After updating to Nextcloud 33, we might consider to bump PHP to v8.4, as this is the [recommended version](https://github.com/nextcloud/server/wiki/Releases-and-PHP-versions) for NC33, while 8.3 is still supported.

stondino00:
Fixes #3487

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -194,8 +194,8 @@ parts:
 
   php:
     plugin: autotools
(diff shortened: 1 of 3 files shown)


pachulo:
Thanks for this @stondino00 ! I've opened https://github.com/nextcloud-snap/nextcloud-snap/pull/3546 instead, which contains just one commit.

PR #3559, opened 2026-07-16
Make MySQL build again to allow for actual development

christianlupus:
Closes #3558 

~~This moves the installation prefix of mysql from `/` (aka $SNAP) to `/mysql` (aka $snap/mysql). Thus, the issue with the empty prefix should be solved.~~
see below

This needs thorough testing as the paths of snap apps and helper scripts might be broken.

PR #3561, opened 2026-07-16
Install rsync as part of the build stage on apache part

### LINKED ISSUE #3560: Building apache fails for missing rsync dependencies
The building of apache as part of the snap fails on my local installation due to missing dependency on `libpopt.so.0`. The call to rsync seems to be related to installing the manual (HTML manual Apache shows by default) to the installation location.

Looking at the snapcraft file, the apache part has no build dependency on rsync, thus, this might break randomly when a different build environment with different default packages is used.

christianlupus:
Fixes #3560

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -135,6 +135,7 @@ parts:
       - libaprutil1-dev
       - libpcre2-dev
       - libssl-dev
+      - rsync
 
     stage-packages:
       - libbrotli1


pachulo:
Does your local build success with this change @christianlupus ?

christianlupus:
I have not tested it explicitly on the latest, rebased branch.
However, I cherry-picked the commit to test #3559 locally. Changing the commit message should not alter the result. I can build this version once, though, to be 100% sure. I am not able to access the machine in the upcoming hours but I can trigger the build this evening.

christianlupus:
OK, I just checked out locally this branch, ran `snapcraft clean --use-lxd` and `snapcraft prime --use-lxd` and it went through smoothly. So, I'd say, the test has passed successfully.

pachulo (approved):

Issue #3587, opened 2026-07-28
Automate dependency handling


christianlupus:
We have some dependency systems in our codebase. For example, there are various Python `requirements.txt` files, the tarballs in the `snapcraft.yaml`, and potentially more. We should/could think of a way to automatically detect required updates and automatically open PRs against these files.

Technically, one could look if Dependabot might be an option to track the versions, however, I am unsure if the `snapcraft.yaml` file format is supported. I am pretty sure that the renovate bot should be able to handle this with a GitHub-based cron job.

pachulo:
I've been checking this and seems like Renovate is indeed the way to go.
~I was thinking about using the [Forking Renovate app](https://github.com/apps/forking-renovate), as it [requires less access to the repository](https://docs.renovatebot.com/security-and-permissions/#global-permissions) and, anyway, we don't really want [the features we would miss by using it](https://docs.renovatebot.com/getting-started/running/#drawbacks).~

I've realized that ours is an advanced use case, because we need to use `baseBranchPatterns` & `postUpgradeTasks/allowedCommands`, so I think that our best option is to run renovate [self-hosted as a GitHub App](https://docs.renovatebot.com/modules/platform/github/#running-as-a-github-app).

Do you see any problems with that approach @kyrofa ?

christianlupus:
I appreciate you found renovate a possible solution. I feared this might be too much machinery...

PR #3634, opened 2026-09-18
CI/CD: configure Renovate for the snap dependencies

pachulo:
OK, I did some research and I think this makes sense. I had my clanker help me with this one, including this very verbose explanation:

> Most snap components are compiled from pinned upstream tarballs listed in `snap/snapcraft.yaml`, so Renovate runs three times a day as a scheduled GitHub Actions workflow (`.github/workflows/renovate.yml`). It opens one pull request per component as soon as a new upstream release is available:
> 
> - Apache httpd, Nextcloud server, PHP, phpredis, APCu, Redis, and MySQL 8.4 LTS, tracked through their upstream GitHub tags.
> 
> Renovate targets `develop` plus the maintained major release branches (`32`, `33`, ...). That list lives in `baseBranchPatterns` in `.github/renovate.json5` and must be updated when a new major release branch is cut. Major upgrades are never proposed automatically: Renovate only opens minor and patch PRs, so every branch (including `develop`) stays on its current Nextcloud/Apache/PHP/MySQL/Redis major.
> PHP and Redis are additionally pinned to the minor line each branch ships (PHP 8.4, Redis 8.2 right now) through `allowedVersions` in `.github/renovate.json5`, because a PHP 8.4 -> 8.5 or Redis 8.2 -> 8.3 jump is a semver-minor bump that the major-disable alone would not block. Major and minor-line bumps are applied manually when the time is right, and the `allowedVersions` pins must be moved along with them.
> 
> GitHub Actions pins are still handled by Dependabot (`.github/dependabot.yml`), and certbot (`src/https/requirements.txt`) is intentionally left out because it follows a separate manual flow.
> 
> Every version bump must come with an updated `source-checksum`. Renovate does this automatically through its `postUpgradeTasks` hook, which runs `tools/update-checksums.sh` before committing the branch. The same script can be run by hand after a manual source bump:
> 
>     $ .github/tools/update-checksums.sh

The workflow authenticates as a [GitHub App rather than a personal access token](https://github.com/renovatebot/github-action#example-with-github-app), so I've already:
1. created an app (`nextcloud-snap-renovate`) with the repository permissions `Contents`, `Pull requests`,
`Issues` and `Commit statuses` set to read/write (`Metadata` is read-only and granted automatically);
2. installed it on the repo;
3. and added its app ID and private key as the `RENOVATE_APP_ID` and `RENOVATE_APP_PRIVATE_KEY` repository secrets.

The workflow exchanges them for a short-lived installation token via [actions/create-github-app-token](https://github.com/actions/create-github-app-token), so no long-lived token is stored.

Also, because scheduled workflows always run from the default branch, this won't run till it hits `master`.

If this works, it should fix https://github.com/nextcloud-snap/nextcloud-snap/issues/3587

@nextcloud-snap/developers I would really like your feedback, specially on the `update-checksums.sh` script.


christianlupus:
I would look into this but on a real screen (no phone) and with some time. This weekend is NC conf and next week is contributor week. I would look into this at some quite minute if you do not mind.

pachulo:
No rush at all @christianlupus , whenever you can!

.github/tools/update-checksums.sh
```diff
@@ -0,0 +1,156 @@
...
+
+# Collect the URLs that changed relative to HEAD (Renovate applies its edits to
+# the working tree before running these tasks, and commits afterwards).
+old_list="$(mktemp)"
+new_list="$(mktemp)"
+map_file="$(mktemp)"
+updated_yaml="$(mktemp)"
+workdir="$(mktemp -d)"
+cleanup() { rm -rf "$old_list" "$new_list" "$map_file" "$updated_yaml" "$workdir"; }
+trap cleanup EXIT
```
christianlupus, line 82: I am not 100% sure if this will cache EVERY possible outcome of the script. E.g I am unsure if you press Ctrl+C or a command fails, what would be the effect. I am pretty sure there is bash/sh/ash discrepancy...

.github/tools/update-checksums.sh
```diff
@@ -0,0 +1,156 @@
...
+checksum_algorithm_for() {
+        local url="$1"
+        awk -v url="$url" '
+                $1 == "source:" && $2 == url { found = 1; next }
+                found && $1 == "source-checksum:" {
+                        split($2, parts, "/")
+                        print parts[1]
+                        exit
+                }
+        ' "$snapcraft_yaml"
```
christianlupus, line 40: This imples that the `source-checksum` is always located after the `source` entry. YAML has no notion of sorting these, so there might be cases where this is broken.

.github/tools/update-checksums.sh
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/env bash
```
christianlupus, line 1: We run this in the GH runner or the local machines, so bash is no issue here, correct?

.github/tools/update-checksums.sh
```diff
@@ -0,0 +1,156 @@
...
+done <<< "$changed"
+
+if [ ! -s "$map_file" ]; then
+        echo "Nothing to update."
+        exit 0
+fi
+
+# Replace the source-checksum that follows each changed source line, keeping the
+# declared algorithm and the original indentation.
+awk -v map_file="$map_file" '
```
christianlupus, line 125: I am not fluent enough in awk to tell if this makes sense of there are side effects or whatever.

I am considering something completely different: would it make sense to make the checking a small python script? I fear, this might be more stable and less hassle (manually writing a YAML parser in Bash/Awk/sed/Grep) to get the file updated? If you want, I can create a python script for that. Just an offer.

.github/renovate.json5
```diff
@@ -0,0 +1,150 @@
+{
+  $schema: "https://docs.renovatebot.com/renovate-schema.json",
+
+  // This is the self-hosted ("global") Renovate configuration used by
```
christianlupus, line 4: I am unsure if comments are allowed here. JSON has normally no comments AFAIK.

.github/renovate.json5
```diff
@@ -0,0 +1,150 @@
...
+
+  // The workflow runs three times a day; open each update immediately instead
+  // of waiting for the next hour (Renovate's default prHourlyLimit is 2).
+  prHourlyLimit: 0,
+
+  // Worst case is one open PR per tracked component per base branch
+  // (7 x 3 = 21); the default prConcurrentLimit of 10 would stall updates on
+  // the remaining branches until PRs merge. Keep this in sync with
+  // baseBranchPatterns and the customManagers below.
+  prConcurrentLimit: 21,
```
christianlupus, line 28: Should we add some safety margin or are we assuming, we will never get close to 21? Especially if there are Python libs changing (see closed PR #3631).

.github/renovate.json5
```diff
@@ -0,0 +1,150 @@
...
+    {
+      matchDepNames: ["php/php-src"],
+      // matchBaseBranches: ["develop", "33"],
+      allowedVersions: "/^8\\.4\\./",
+    },
+    {
+      matchDepNames: ["redis/redis"],
+      allowedVersions: "/^8\\.2\\./",
+    },
+  ],
```
christianlupus, line 52: Is this needed/good?
We might know not to blindly merge a new PHP or redis version. But still, it would be nice to get a _notificification_ already prepared, no? All deps should be in isolated PRs if I am not misled. So, we can decide which PRs to merge and which better not.
WDYT?

christianlupus (approved):
I have just now looked through these files (and being half-asleep might make my thoughts a bit sluggish). Nevertheless, I tried to work my way through the files and put some comments there. Generally, I would approve the changes, but I'd like to see the comments settled.

.github/tools/update-checksums.sh
pachulo: True, we should take that into account! I guess we should use something like [yq](https://github.com/mikefarah/yq) if we stick with bash, or [ruamel.yaml](https://pypi.org/project/ruamel.yaml/) if we implement this part in Python.

.github/tools/update-checksums.sh
pachulo: My bash fu is weak, so I do neither know if it would cover everything...

.github/tools/update-checksums.sh
pachulo: True

.github/tools/update-checksums.sh
pachulo: > I am considering something completely different: would it make sense to make the checking a small python script? I fear, this might be more stable and less hassle (manually writing a YAML parser in Bash/Awk/sed/Grep) to get the file updated? If you want, I can create a python script for that. Just an offer.

If you could it, that would be awesome.
Thanks!

.github/renovate.json5
pachulo: I have no idea, to be honest, but I think that the clanker validated the config with https://docs.renovatebot.com/config-validation/

.github/renovate.json5
pachulo: This is taking into account the dependencies that we want to tackle with Renovate, which are the 7 things that we compile from source code in the snap.

I would not worry about Python dependencies right now and anyway, if we wanted to Dependabot is ready to open PR for Python dependencies which are not security related.

.github/renovate.json5
pachulo: Yes, IMHO  is needed. I prefer to have to manually update to new "branches" of Redis & PHP.
Besides, PHP upgrades need more tweaking that just updating the snapcraft part, [see here](https://github.com/nextcloud-snap/nextcloud-snap/commit/4ea57189b5a7eab90bb7c6b1944841c6833fecd6), so they wouldn't work anyway.

.github/tools/update-checksums.sh
christianlupus: Give me a chance to do this tomorrow. This might probably be the easiest if I am in front of a monitor and a keyboard. 😉

.github/renovate.json5
christianlupus: I was more thinking in the direction "to get a notification there was something to keep in mind".

But I am okay with it as it is. So, you can also resolve the conversation.

pachulo: > I was more thinking in the direction "to get a notification there was something to keep in mind".

That would add a marginal value, as we focus on the versions supported (and recommended) in the nextcloud docs. And those I manually monitor.

